PR #3, opened 2018-12-05
revert blas_openblas feature patches

minrk:
since only the main channel (no labels) is hotfixed, this results in inconsistent installation behavior for the gcc7 channel where blas still has `track_features`.

As it is now, new builds on main *must* remove the `blas_openblas` feature to be installed, and builds on gcc7 *must not* remove the `blas_openblas` feature, which is likely to cause problems.

This should be temporary, as when the gcc7 channel is promoted to main, the hotfix should be able to go back in, or we can wait for one of the other blas 2.x proposals to remove track_features.

Checklist
* [x] Used a fork of the feedstock to propose changes
* [x] Bumped the build number (if the version is unchanged)
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -258,16 +258,19 @@ def _patch_repodata(repodata, subdir):
             depends = record["depends"]
             depends.append("_r-mutex 1.* anacondar_1")
             instructions["packages"][fn]["depends"] = depends
-        if record_name == "blas" and record["track_features"] == "blas_openblas":
-            instructions["packages"][fn]["track_features"] = None
-        if "features" in record:
-            if "blas_openblas" in record["features"]:
-                # remove blas_openblas feature
-                instructions["packages"][fn]["features"] = _extract_feature(record, "blas_openblas")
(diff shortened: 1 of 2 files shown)


minrk:
fwiw, this is a little more time sensitive/serious than I thought: It's not just that the latest builds of packages with `blas_openblas` aren't propagating to users. They can't even be built. This means, e.g. that all builds of numpy, scipy, suitesparse, etc. will fail until this is resolved. Every package that has `blas_openblas` feature cannot submit a new version or build.

minrk:
Thanks! @soapy1 when/how does this get deployed to the mirror? Is it automatic?

minrk:
Appears to have been automatic (or a human was very quick!). The previously not-working installs are working now, at least when I test on mac and linux.

msarahan:
It is automatic. The mirroring script picks up the latest version of this package each time it gets run.

PR #14, opened 2019-03-18
Fix libspatialindex pin

ocefpaf:
This is probably not correct but I wanted to send something to start a conversation.

The problem is, we never pinned `libspatialindex` in the past b/c there weren't new versions of it, the project seemed dead. However, thanks to @hobu, the project is back and we have new versions! So, before merging [`libspatialindex 1.9.0`](https://github.com/conda-forge/libspatialindex-feedstock/pull/18), we need to fix the unpinned packages we have to make sure they won't pull version `1.9.0` and break things.

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -279,6 +279,10 @@ def _gen_new_index(repodata, subdir):
         if "ntl" in deps and record_name != "sage":
             _rename_dependency(fn, record, "ntl", "ntl 10.3.0")
 
+        deps = record.get("depends", ())
+        if "libspatialindex" in deps:
+            _rename_dependency(fn, record, "libspatialindex", "libspatialindex 1.8")
+
         # FIXME: disable patching-out blas_openblas feature
         # because hotfixes are not applied to gcc7 label
         # causing inconsistent behavior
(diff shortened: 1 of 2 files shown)


recipe/gen_patch_json.py
```diff
@@ -279,6 +279,10 @@ def _gen_new_index(repodata, subdir):
         if "ntl" in deps and record_name != "sage":
             _rename_dependency(fn, record, "ntl", "ntl 10.3.0")
 
+        deps = record.get("depends", ())
+        if "libspatialindex" in deps:
+            _rename_dependency(fn, record, "libspatialindex", "libspatialindex 1.8")
```
isuruf: I don't think this works. (The ntl code above is wrong too)

ocefpaf: :smile: 

No idea what is supposed to work here though.

soapy1: After a bit of testing, it seems like this will only apply a pin to libspatialindex 1.8 if the package is unpinned. So, in the future, if you want to be able to use libspatialindex 1.9 I think this will still be ok, provided that the dependency is pinned to be something like `> 1.8`.
I'm not entirely sure about pinning directly to 1.8 though, is there any case where a package will want libspatialindex < 1.8?

ocefpaf: > I'm not entirely sure about pinning directly to 1.8 though, is there any case where a package will want libspatialindex < 1.8?

Nope. `libspatialindex` was not updated in ages so we have mostly `1.8.5` and `1.9.0` only available.

Thanks for reviewing this @soapy1! Is it good to be merged? I'm not sure what are @isuruf's concerns here.

isuruf: If we had `libspatialindex 1.9` wouldn't this change it to `libspatialindex 1.8 1.8`

ocefpaf: Unless those `deps` are only unpinned then I guess it would be `libspatialindex 1.8 1.9` :-|

ocefpaf:
> lgtm. After this is merged there will be about a 20 min delay before it gets applied to repodata because of how mirroring is working

Thanks! Let's merge it then.

PR #15, opened 2019-03-21
Revert "Fix libspatialindex pin"

ocefpaf:
Reverts conda-forge/conda-forge-repodata-patches-feedstock#14

Ping @soapy1


recipe/gen_patch_json.py
```diff
@@ -357,35 +353,6 @@ def _fix_libgfortran(fn, record):
             depends[dep_idx] = "libgfortran >=4.0.0,<5.0.0.a0"
             record['depends'] = depends
 
-def _fix_libnetcdf(fn, record):
```
ocefpaf: Jeez this should not be here at all :-/

PR #17, opened 2019-08-12
fix constraint on conda in conda-build 3.18 from 4.3 to 4.5

msarahan:
This also changes some of the if's to be elifs, for a slight increase in efficiency.


hmaarrfk:
Sorry for letting this slip, but is this still relevant 2 years later

jezdez:
IDK, I don't think we still support these versions

PR #32, opened 2020-02-26
hotfix distributed <2.11.0 to bound msgpack-python

jjhelmus:
Adjust the requirements of distributed <2.11.0 to bound msgpack-python
to less than 1.0.0

Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -455,6 +455,15 @@ def _gen_new_index(repodata, subdir):
             if new_constrains != record.get('constrains', ()):
                 record['constrains'] = new_constrains
 
+        # distributed <2.11.0 does not work with msgpack-python >=1.0
+        # newer versions of distributed require at least msgpack-python >=0.6.0
+        # so we can fix cases where msgpack-python is unbounded
+        # https://github.com/conda-forge/distributed-feedstock/pull/114
+        if record_name == 'distributed'
+            if 'msgpack-python' in record['depends']:
+                i = record['depends'].index('msgpack-python')
+                record['depends'][i] = 'msgpack-python <1.0.0'
+
         # fix deps with wrong names
         if record_name in proj4_fixes:
             _rename_dependency(fn, record, "proj.4", "proj4")
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -459,7 +459,7 @@ def _gen_new_index(repodata, subdir):
         # newer versions of distributed require at least msgpack-python >=0.6.0
         # so we can fix cases where msgpack-python is unbounded
         # https://github.com/conda-forge/distributed-feedstock/pull/114
-        if record_name == 'distributed'
+        if record_name == 'distributed':
             if 'msgpack-python' in record['depends']:
                 i = record['depends'].index('msgpack-python')
                 record['depends'][i] = 'msgpack-python <1.0.0'
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,13 +1,13 @@
 package:
   name: conda-forge-repodata-patches
-  version: 20200211
+  version: 20200226
 
 source:
   path: .
 
 build:
   noarch: generic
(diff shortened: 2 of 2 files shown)


beckermr:
Ahhhhhhh!

I think this one explains a pin I had to put somewhere else.

recipe/gen_patch_json.py
```diff
@@ -455,6 +455,15 @@ def _gen_new_index(repodata, subdir):
...
                 record['constrains'] = new_constrains
 
+        # distributed <2.11.0 does not work with msgpack-python >=1.0
+        # newer versions of distributed require at least msgpack-python >=0.6.0
+        # so we can fix cases where msgpack-python is unbounded
+        # https://github.com/conda-forge/distributed-feedstock/pull/114
+        if record_name == 'distributed'
+            if 'msgpack-python' in record['depends']:
+                i = record['depends'].index('msgpack-python')
+                record['depends'][i] = 'msgpack-python <1.0.0'
```
beckermr: do you also want `>=0.6.0` here too?

jjhelmus: I'm not sure, the minimum version for 2.11.0 is 0.6.0 but ealier version may have been compatible with older msgpack-python versions.

@jakirkham do you have insight into this?

jjhelmus:
@jrbourbeau was interested in this as well

jjhelmus:
@beckermr Thanks for the fix, apparently writing code during a meeting is not a good idea :smile:.

recipe/gen_patch_json.py
beckermr: Huh? I mean the minimum version of msgpack of 0.6.0.

jrbourbeau: > So I don't think we need this for older versions

Agreed. Older versions of `distributed` work fine with pre-0.6.0 versions of msgpack. We set the minimum version of msgpack to 0.6.0 because that's when a certain keyword argument was added that we needed to fix the msgpack 1.0 compatibility issues

beckermr: ohhhhhh I see. OK. Then LGTM!

jakirkham: > Huh? I mean the minimum version of msgpack of 0.6.0.

Sorry if that's confusing. The risk of reviewing PRs during a conference. 😉 Hopefully James clarified this, but please let us know if you have more questions. 🙂

beckermr:
Woops! I got trigger happy on the merge button here! Sorry @jakirkham !

beckermr:
Ack sorry all. This code will patch all versions of distributed, not just less 2.11.0. We will need to back it out and change it to patch only certain versions.

jjhelmus:
> Ack sorry all. This code will patch all versions of distributed, not just less 2.11.0

The `if 'msgpack-python' in record['depends']:` line will only match below 2.11.0, newer version of distributed have `msgpack-python >=0.6.0` in the depends list.

jakirkham:
> Woops! I got trigger happy on the merge button here! Sorry @jakirkham !

Nothing to be sorry about. Thanks for merging 😄

recipe/gen_patch_json.py
Carreau: Questions for hotfixes: Once a new version of distributed compatible with msgpack 1.0 is released, does that have any chance to re-pin it to <1.0 when re-ran ?

isuruf: Because of line 463, this fixing is only applied to distributed versions with a dependency of msgpack-python without any version constraints. If there's a new version of distributed with a requirement like
```
- msgpack-python
```
instead of
```
- msgpack-python >=0.6
```
then, yes.

Carreau: Ah ! Yes, thank, for whatever reason I thought that `record['depends']` would be a dict with `msgpack-python` a key. Makes sens.

PR #35, opened 2020-03-15
Add pin for Pythia 8

chrisburr:
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Ensured the license file is being packaged.

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -515,6 +515,11 @@ def _gen_new_index(repodata, subdir):
             i = record['depends'].index('abseil-cpp')
             record['depends'][i] = 'abseil-cpp =20190808'
 
+        # Filter by timestamp as pythia8 also contains python bindings that shouldn't be pinned
+        if 'pythia8' in deps and record.get('timestamp', 0) < 1584264455759:
+            i = record['depends'].index('pythia8')
+            record['depends'][i] = 'pythia8 >=8.240,<8.300.0a0'
+
         # remove features for openjdk and rb2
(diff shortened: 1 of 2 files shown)


chrisburr:

chrisburr:
@conda-forge/core Ready for review/merge

recipe/gen_patch_json.py
```diff
@@ -515,6 +515,11 @@ def _gen_new_index(repodata, subdir):
             i = record['depends'].index('abseil-cpp')
             record['depends'][i] = 'abseil-cpp =20190808'
 
+        # Filter by timestamp as pythia8 also contains python bindings that shouldn't be pinned
+        if 'pythia8' in deps and record.get('timestamp', 0) < 1584264455759:
```
xhochy: Can you explain a bit more about this magic timestamp?

chrisburr: This is under the assumption that these patches apply to all conda-forge packages both present and future. I haven't seen anything confirming this but it appears to be the case?

The timestamp is just from the newest package in conda-forge at the time of me committing this line I'm confident everything currently in conda-forge won't be negatively affected by this.

In future packages might be added which depend on `pythia8` as a python library and these shouldn't inherit a pin for some random old version which had bad metadata.

xhochy: Sounds reasonable 👍 but I would need someone else here to approve as I cannot confirm all assumptions.

PR #36, opened 2020-03-19
Hotfix license_family

isuruf:
Checklist
* [ ] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/get_license_family.py
+++ b/recipe/get_license_family.py
@@ -1,0 +1,38 @@
+import license_expression
+
+mapping = {
+    "GPL-2.0-or-later": "GPL",
+    "BSD-3-Clause": "BSD",
+    "MIT": "MIT",
+}
+
+precedence = {
+    ("BSD", "GPL"): "GPL",
+    ("GPL", "MIT"): "GPL",
+}
+
+def get_license_family(license):
+    licensing = license_expression.Licensing()
+    family = None
+    try:
+        parsed_licenses_with_exception = licensing.license_symbols(
+            license.strip(), decompose=False
+        )
+        for l in parsed_licenses_with_exception:
+            if isinstance(l, license_expression.LicenseWithExceptionSymbol):
+                license = l.license_symbol.key
+            else:
+                license = l.key
+            family_new = mapping.get(license, None)
+            if not family_new:
+                return None
+            if family is None:
+                family = family_new
+                continue
+            pair = tuple(sorted([family_new, family]))
+            family = precendence.get(pair, None)
+            if not family:
+                return None
+    except license_expression.ExpressionError:
+        return None
+    return family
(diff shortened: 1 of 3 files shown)


recipe/get_license_family.py
```diff
@@ -0,0 +1,38 @@
+import license_expression
+
+mapping = {
+    "GPL-2.0-or-later": "GPL",
+    "BSD-3-Clause": "BSD",
+    "MIT": "MIT",
+}
```
beckermr: Do we need many more entries here, or did you confirm these are all of the ones that we can fix and/or need fixing?

recipe/get_license_family.py
```diff
@@ -0,0 +1,38 @@
+import license_expression
```
beckermr: Smithy depends on this so the import is fine, but you probably want to add it to the deps directly here so that if we ever remove it from smithy, this feedstock will still build.

beckermr: looks like smithy is not installed here as I had assumed so you need to list the dep.

recipe/get_license_family.py
isuruf: We do. I don't have time to go through all of them, but hope others can help me.

beckermr: Ahhh got it. Many of these can be generated via regex. Have you been able to get a unique list? This repo may help: https://github.com/regro/libcfgraph

CJ-Wright: It would be great if we could centralize this knowledge, I'm thinking of building a mini migrator to fix this in the recipes themselves and this would be super helpful.

beckermr: +1 Whatever mapping is generated can be stored as yaml or json and used in both spots. It doesn't have to be dynamic in that eventually we will catch most of everything.

isuruf: I'll only add `SPDX->license_family` mapping here. For the bot you need `license->SPDX` mapping.

isuruf:
@beckermr, this is ready now.

jakirkham:
Could you please summarize this change in the OP? I think I get the idea, but it would be helpful for overall clarity 🙂

recipe/get_license_family.py
```diff
@@ -0,0 +1,57 @@
...
+    "MPL-1.1": "MOZILLA",
+    "MPL-2.0": "MOZILLA",
+    "MPL-2.0-no-copyleft-exception": "MOZILLA",
+    "LicenseRef-Proprietary": "PROPRIETARY",
+}
+
+precedence = {
+    ("BSD", "GPL"): "GPL",
+    ("GPL", "MIT"): "GPL",
+}
```
beckermr: Do we need all families here?

isuruf: We do, but we don't have many licenses with multiple ones. I'd rather add this if/when someone is interested.

isuruf:
Thanks @beckermr for the review

PR #38, opened 2020-03-25
Pin old versions of HoloViews to bokeh<2.0

philippjfr:
So I'm not entirely happy with this patch but I had to use some heuristic to detect HoloViews versions <1.13.0. Since we only just introduced the hard dependency on pandas that was the best heuristic I could find. Happy to hear better suggestions, e.g. using LooseVersion to compare the actual HoloViews version, I've found that can be a little bit brittle at times though.

Also I guess I may have missed some intermediate pins, I know early versions did not pin bokeh at all (yikes) and more recent versions pinned "bokeh >=1.1.0" but there's probably a bunch of intermediate versions I missed since there's so many of them.

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -471,6 +471,20 @@ def _gen_new_index(repodata, subdir):
                 i = record['depends'].index('msgpack-python')
                 record['depends'][i] = 'msgpack-python <1.0.0'
 
+        # HoloViews <1.13.0 does not work with bokeh >= 2.0
+        # Newer versions do allow it but have the same pin as older
+        # versions, therefore we use the fact that 1.13.0 is the first
+        # version to require pandas to detect >1.13.0
+        if record_name == 'holoviews':
+            if 'bokeh' in record['depends']:
+                i = record['depends'].index('bokeh')
+                record['depends'][i] = 'bokeh <2.0.0'
+            elif ('bokeh >=1.1.0' in record['depends'] and
+                  not any('pandas' in d for d in record['depends'])):
+                i = record['depends'].index('bokeh >=1.1.0')
+                current = record['depends'][i]
+                record['depends'][i] = current+',<2.0.0'
+
         # fix deps with wrong names
         if record_name in proj4_fixes:
             _rename_dependency(fn, record, "proj.4", "proj4")


philippjfr:
Probably going to keep thinking about this, maybe a better approach would be to append `,<2.0.0` to all bokeh version pins (dropping the comma if the pin is simply `bokeh`).

recipe/gen_patch_json.py
```diff
@@ -471,6 +471,20 @@ def _gen_new_index(repodata, subdir):
...
+        # version to require pandas to detect >1.13.0
+        if record_name == 'holoviews':
+            if 'bokeh' in record['depends']:
+                i = record['depends'].index('bokeh')
+                record['depends'][i] = 'bokeh <2.0.0'
+            elif ('bokeh >=1.1.0' in record['depends'] and
+                  not any('pandas' in d for d in record['depends'])):
+                i = record['depends'].index('bokeh >=1.1.0')
+                current = record['depends'][i]
+                record['depends'][i] = current+',<2.0.0'
```
jakirkham:
```suggestion
                record['depends'][i] = current + ',<2.0.0'
```

jakirkham:
@conda-forge/core, could someone else please take a look too? 🙂

philippjfr:
I think I want to do the same for older versions of Panel. Any objections to adding that here or would you prefer a new PR?

jakirkham:
Not from me. Maybe update the title and OP to reflect the change of scope?

jjhelmus:
I do not think this logic is sufficient to correct all the `holoview` packages currently in conda-forge.  A number of the earlier version have a minimum bound on boken which would not be correctly with the current matching scheme. 

The current bokeh requirements for the holoview packages are:
```
$ conda search --override-channels -c conda-forge holoviews --info | grep bokeh | sort | uniq 
  - bokeh
  - bokeh >=0.12.14,<=0.12.16
  - bokeh >=0.12.15
  - bokeh >=0.12.15,<=0.12.16
  - bokeh >=0.12.15,<=0.13.0
  - bokeh >=0.12.15,<1.1.0
  - bokeh >=1.0.0,<1.1.0
  - bokeh >=1.1.0
  - bokeh >=1.1.0,<2
  - bokeh >=1.1.0,<2.0.1
```

Those with a dependency of `bokeh >=0.12.15` and `- bokeh >=1.1.0` would have any upper limit on their bokeh version.

owenlamont:
Interesting, I need to make a near identical change to the jupyterlab-nvdashboard package - it has earlier releases missing a bokeh <2 requirement. I hope I can re-use some similar logic for my problem.

hmaarrfk:
no response for a few months, i'll close. please re-open if still an issue.

PR #39, opened 2020-03-30
Patch libnetcdf pin

ocefpaf:
Ping @isuruf

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -498,7 +498,7 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("gf2x") for dep in deps):
             _pin_stricter(fn, record, "gf2x", "x.x")
 
-        if any(dep.startswith("libnetcdf") for dep in deps):
+        if any(dep.startswith("libnetcdf >=4.7.3") for dep in deps):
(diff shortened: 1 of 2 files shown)


recipe/gen_patch_json.py
```diff
@@ -498,6 +498,9 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("gf2x") for dep in deps):
             _pin_stricter(fn, record, "gf2x", "x.x")
 
+        if any(dep.startswith("libnetcdf >=4.7.3") for dep in deps):
```
isuruf: @ocefpaf, is it okay to fix only packages built with `4.7.3`?

ocefpaf: Maybe >=4.7.0 to be safe? B/c old build with 4.7.{0,1,2} will install 4.7.4 too, right?

ocefpaf: What do you think? Does that makes sense?

isuruf: There are no builds with 4.7.0, 4.7.2. `4.7.1` is pinned to `x.x.x`

ocefpaf: Great. So let's get this one merge then!

ocefpaf:
Thanks @isuruf!

PR #74, opened 2020-07-29
Patch repodata to add sysroot conflict for root_base and cling

chrisburr:
See https://github.com/conda-forge/clangdev-feedstock/pull/104#issuecomment-665607590 for justification.

cc @beckermr 


recipe/gen_patch_json.py
```diff
@@ -666,7 +666,7 @@ def _gen_new_index(repodata, subdir):
             subdir in ["linux-64", "linux-aarch64", "linux-ppc64le"]
             and record_name in [
                 "gcc_" + subdir, "gxx_" + subdir, "gfortran_" + subdir,
-                "binutils_" + subdir, "gcc_bootstrap_" + subdir]
+                "binutils_" + subdir, "gcc_bootstrap_" + subdir, "root_base", "cling"]
```
beckermr, line 669: Can you add some comments on why these two are in this block of code? The current comments don't mention them.

chrisburr:
That's fine, I just checked with `strace` and they do have a real runtime dependency on the contents of `sysroot_linux-64`.

I'm running more local tests at the moment and I'll then make the necessary PRs. Thanks for the rapid help! 🍰

beckermr:
When you are ready to merge this, let us know!

chrisburr:
I think this can go in now as my tests aren't going to change it.

Issue #80, opened 2020-08-15
Fix libiconv repodata


isuruf:
1.15 is compatible with 1.16

cc @scopatz

PR #82, opened 2020-08-17
libiconv 1.15 and 1.16 are compatible

isuruf:
Fixes https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/issues/80

cc @scopatz

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -539,6 +539,9 @@ def _gen_new_index(repodata, subdir):
         if "ntl" in deps and record_name != "sage":
             _rename_dependency(fn, record, "ntl", "ntl 10.3.0")
 
+        if "libiconv >=1.15,<1.16.0a0" in deps:
+            _pin_looser(fn, record, "libiconv", upper_bound="1.17.0")
+
         if 're2' in deps and record.get('timestamp', 0) < 1588349339243:
             _rename_dependency(fn, record, "re2", "re2 <2020.05.01")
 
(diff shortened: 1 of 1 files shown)


beckermr:
Do we need to adjust the iconv run exports and pinning expression to pin only to the major version?

isuruf:
No, because 1.17 might not be compatible.

beckermr:
I see. So to summarize we'd like to have a system that does more fine-grained ABI compatibility so that we can avoid shipping a version every now and then.

isuruf:
I didn't understand your comment. Summary is that for packages that doesn't use semver, we have to pin strict and loosen when we know that the versions are compatible.

beckermr:
Right OFC. But I'd be nice to automate the loosening of the pins and if we did that, it should happen on a case-by-case basis for each new version.

isuruf:
Sure. It would be nice to have.

Merge?

beckermr:
ofc merge away!

CJ-Wright:
Can we get that up as an issue though? Esp for the new spec `x.x` might not be very descriptive. It might be better to explicitly define the min and max compatible versions. This would cover most cases, at least until packages return to a previous ABI. In this case that would have been a change on the feedstock, when it got the version bump it would have bumped the max compatible version.

isuruf:
>  In this case that would have been a change on the feedstock, when it got the version bump it would have bumped the max compatible version.

No, metadata in downstream packages need to be fixed.

PR #86, opened 2020-09-09
include patches for osx-arm64

jjhelmus:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -34,7 +34,7 @@ test:
     - license-expression
     - pytest
   commands:
-    {% for subdir in ("noarch", "linux-64", "linux-armv7l", "linux-ppc64le", "osx-64", "win-32", "win-64") %}
+    {% for subdir in ("noarch", "linux-64", "linux-armv7l", "linux-ppc64le", "osx-64", "osx-64", "win-32", "win-64") %}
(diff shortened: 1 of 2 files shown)


isuruf:
Thanks @jjhelmus

beckermr:
Thank you @jjhelmus!

recipe/meta.yaml
```diff
@@ -34,7 +34,7 @@ test:
     - license-expression
     - pytest
   commands:
-    {% for subdir in ("noarch", "linux-64", "linux-armv7l", "linux-ppc64le", "osx-64", "win-32", "win-64") %}
+    {% for subdir in ("noarch", "linux-64", "linux-armv7l", "linux-ppc64le", "osx-64", "osx-64", "win-32", "win-64") %}
```
jakirkham, line 37: Should the second `"osx-64"` here actually be `"osx-arm64"`?

jjhelmus: Oops.  Yes you are correct there should be a `osx-arm64` in that list. Copy and paste typo.

jjhelmus: Should mention this does not cause any issues in the package since this is in the test.  The presence of the `osx-64` patch get checked twice (so it is REALLY there).

jakirkham: Ah ok. Thanks for clarifying and fixing 😄

PR #110, opened 2021-01-15
Loose pin for openmpi and mpich

leofang:
Checklist

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -654,6 +654,12 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("zstd >=1.4") for dep in deps):
             _pin_looser(fn, record, "zstd", max_pin="x.x")
 
+        if "openmpi >=4.1" in deps:
+            _pin_looser(fn, record, "openmpi", upper_bound="5.0")
+
+        if "mpich >=3.3" in deps:
+            _pin_looser(fn, record, "mpich", upper_bound="4.0")
+
         _replace_pin('libunwind >=1.2.1,<1.3.0a0', 'libunwind >=1.2.1,<2.0.0a0', deps, record)
         _replace_pin('snappy >=1.1.7,<1.1.8.0a0', 'snappy >=1.1.7,<2.0.0.0a0', deps, record)
         _replace_pin('ncurses >=6.1,<6.2.0a0', 'ncurses >=6.1,<6.3.0a0', deps, record)
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -654,10 +654,10 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("zstd >=1.4") for dep in deps):
             _pin_looser(fn, record, "zstd", max_pin="x.x")
 
-        if "openmpi >=4.1" in deps:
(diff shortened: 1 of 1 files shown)


leofang:
cc: @isuruf @beckermr

recipe/gen_patch_json.py
```diff
@@ -654,6 +654,12 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("zstd >=1.4") for dep in deps):
             _pin_looser(fn, record, "zstd", max_pin="x.x")
 
+        if any(dep.startswith("openmpi >=4.1") for dep in deps):
```
isuruf, line 657:
```suggestion
        if any(dep.startswith("openmpi >=4.0") for dep in deps):
```

leofang: 👍

beckermr (changes requested):
Don't we want to adjust anything matching 4.x for openmpi and have no upper bound  for mpich.

leofang:
Why no upper bound for mpich? Is its ABI always stable?!

beckermr:
@isuruf told me it's abi is stable across major versions.

beckermr:
Wait a second. Abi lab says it changed soname on 3.4. Does that not matter?

isuruf:
> @isuruf told me it's abi is stable across major versions.

Sorry, I was wrong, but mpich hasn't bumped the major version in years.

isuruf:
This looks good to me, but we need to fix the `run_exports` of `mpich` first.

leofang:
Hmm 🤔 But it also says 100% compatible... I don't get it either

leofang:
Oh yeah nvm. It's not even beta 😂

recipe/gen_patch_json.py
```diff
@@ -654,6 +654,12 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("zstd >=1.4") for dep in deps):
             _pin_looser(fn, record, "zstd", max_pin="x.x")
 
+        if any(dep.startswith("openmpi >=4.0") for dep in deps):
```
leofang, line 657: @isuruf Wouldn't it be better if we just say 'openmpi>=4' here?

isuruf: Nope. We don't want to patch newer packages. Repodata patches should apply to  older ones and not new ones.

leofang: But what happens to openmpi 4.1.0?

leofang:
> This looks good to me, but we need to fix the `run_exports` of `mpich` first.

https://github.com/conda-forge/mpich-feedstock/pull/54 is in.

leofang:
@conda-forge/core this should be ready?

leofang:
Thanks, @isuruf. How long does it take to be in effect?

isuruf:
1 CDN sync (~ 1 hour)

leofang:
OK thanks. I will check it later to see if https://github.com/conda-forge/mpi4py-feedstock/issues/36 is fixed.

leofang:
Looks like we're good!

beckermr:
I am seeing pins from 3.4 to 3.5 in mpi4py. I think we need to patch again.

beckermr:
also this binary is not patched: osx-64/hdf5-1.10.5-mpi_mpich_hd288fd1_1014.tar.bz2

PR #124, opened 2021-02-18
Patch dependency on tbb

AlexVeprev:
This PR is created after the following discussions:

- https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/1227
- https://github.com/conda-forge/tbb-feedstock/pull/75

It is required for safe upgrade of TBB to oneTBB 2021 that is incompatible with previous versions: https://github.com/conda-forge/tbb-feedstock/pull/78

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -717,7 +717,7 @@ def _gen_new_index(repodata, subdir):
         # TBB 2021 (oneTBB 2021) is incompatible with previous releases.
         if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614261110000:
             for i, dep in enumerate(deps):
-                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2017, 2020)) or dep.startswith("tbb >=4.4"):
+                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2017, 2021)) or dep.startswith("tbb >=4.4"):
                     deps.append("tbb <2021.0.0a0")
                     break
 
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -717,9 +717,7 @@ def _gen_new_index(repodata, subdir):
         # TBB 2021 (oneTBB 2021) is incompatible with previous releases.
         if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614285451000:
             for i, dep in enumerate(deps):
-                if dep.split(" ")[0] != "tbb":
-                    continue
-                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2016, 2021)):
+                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2017, 2020)) or dep.startswith("tbb >=4.4"):
                     deps.append("tbb <2021.a0")
                     break
 
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -718,7 +718,7 @@ def _gen_new_index(repodata, subdir):
         if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614261110000:
             for i, dep in enumerate(deps):
                 if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2017, 2020)) or dep.startswith("tbb >=4.4"):
-                    deps.append("tbb <2021.a0")
+                    deps.append("tbb <2021.0.0a0")
                     break
 
         _replace_pin('libunwind >=1.2.1,<1.3.0a0', 'libunwind >=1.2.1,<2.0.0a0', deps, record)
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -719,13 +719,9 @@ def _gen_new_index(repodata, subdir):
             for i, dep in enumerate(deps):
                 if dep.split(" ")[0] != "tbb":
                     continue
-                if "<" in dep:
+                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2016, 2021)):
+                    deps.append("tbb <2021.a0")
                     break
-
-                if " " in record['depends'][i]:
-                    record['depends'][i] = record['depends'][i] + ',<2021a0'
-                else:
-                    _rename_dependency(fn, record, "tbb", "tbb <2021a0")
 
         _replace_pin('libunwind >=1.2.1,<1.3.0a0', 'libunwind >=1.2.1,<2.0.0a0', deps, record)
         _replace_pin('snappy >=1.1.7,<1.1.8.0a0', 'snappy >=1.1.7,<2.0.0.0a0', deps, record)
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -715,7 +715,7 @@ def _gen_new_index(repodata, subdir):
             _pin_looser(fn, record, "mpich", upper_bound="4.0")
 
         # TBB 2021 (oneTBB 2021) is incompatible with previous releases.
-        if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614285451000:
+        if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614261110000:
             for i, dep in enumerate(deps):
                 if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2017, 2020)) or dep.startswith("tbb >=4.4"):
                     deps.append("tbb <2021.a0")
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -714,6 +714,19 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("mpich >=3.4.1,") for dep in deps):
             _pin_looser(fn, record, "mpich", upper_bound="4.0")
 
+        # TBB 2021 (oneTBB 2021) is incompatible with previous releases.
+        if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614285451000:
+            for i, dep in enumerate(deps):
+                if dep.split(" ")[0] != "tbb":
+                    continue
+                if "<" in dep:
+                    break
+
+                if " " in record['depends'][i]:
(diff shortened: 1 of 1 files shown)


recipe/gen_patch_json.py
```diff
@@ -714,6 +714,19 @@ def _gen_new_index(repodata, subdir):
...
+            for i, dep in enumerate(deps):
+                if dep.split(" ")[0] != "tbb":
+                    continue
+                if "<" in dep:
+                    break
+
+                if " " in record['depends'][i]:
+                    record['depends'][i] = record['depends'][i] + ',<2021a0'
+                else:
+                    _rename_dependency(fn, record, "tbb", "tbb <2021a0")
```
isuruf, line 728: this would break if the dependency is `tbb 2020.*` or if it is strictly pinned.
I would suggest the following,
```suggestion
                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2016, 2021)):
                    deps.append("tbb <2021.a0")
                    break
```

AlexVeprev: @isuruf, thank you!

I've applied your suggestion and updated it a bit.
Could you please confirm that it will be fine if a package has 2 "tbb"s under "depends": the first with lower bound (`>=`) or with no bound, and another one with upper bound `<2021.a0`?
E.g.
```
-    "tbb >=2019.0"
+    "tbb >=2019.0",
+    "tbb <2021.a0"
```
While one may expect something like that:
`tbb >=2019.0` --> `tbb >=2019.0,<2021.a0`
`tbb` --> `tbb <2021.a0`

AlexVeprev: Also I have some doubts about `.` before `a0` in your suggestion: is it really needed? I'm not familiar with such versioning details but I see regex in this file:
```python
cb_pin_regex = re.compile(r"^>=(?P<lower>\d(\.\d+)*a?),<(?P<upper>\d(\.\d+)*)a0$")
```
that is used in `_pin_stricter` and `_pin_looser` and it assumes `a0` suffix with no dot before it. BTW this regex won't match for tbb bounds anyway because one-digit major version is expected.

chrisburr: >  if a package has 2 "tbb"s under "depends

This is fine, the solver will look for a solution that matches all constraints.

(other people are better qualified than me for the other points)

AlexVeprev:
All the comments seems to be addressed, @isuruf, @chrisburr, should I do something else to merge the PR?

recipe/gen_patch_json.py
```diff
@@ -714,6 +714,13 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("mpich >=3.4.1,") for dep in deps):
             _pin_looser(fn, record, "mpich", upper_bound="4.0")
 
+        # TBB 2021 (oneTBB 2021) is incompatible with previous releases.
+        if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614285451000:
```
chrisburr, line 718: I guess this timestamp should be updated once https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/124 is deployed?

AlexVeprev: Oh, yes, did you mean deployment of https://github.com/conda-forge/tbb-feedstock/pull/78 ?

AlexVeprev: I wonder if I should set some timestamp in future and then align it after deployment if it is done sooner

chrisburr: Oops 😅 I meant to link https://github.com/conda-forge/tbb-feedstock/pull/79 with the view that any builds from now onwards should already have the correct pinning

AlexVeprev: Got it, https://github.com/conda-forge/tbb-feedstock/pull/79 is merged, so I've updated the timestamp to Thu Feb 25 2021 13:51:50 UTC

chrisburr: Perfect, can you post the output of running `show_diff.py`

chrisburr: Also 1b of https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/1227#issuecomment-781493615 can be added in parallel to this PR if you have a moment.

AlexVeprev: `python show_diff.py > tbb-patched-version.diff.log`: [tbb-patched-version.diff.log](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/files/6044331/tbb-patched-version.diff.log)

Regarding 1b: I don't fully understand the idea of pinning in conda-forge-pinning-feedstock, but I've just created https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1241 to start, feel free to comment, reject or update it.

AlexVeprev: BTW I've updated 2021.a0 --> 2021.0.0a0 to align with conda-forge/tbb-feedstock#79

recipe/gen_patch_json.py
```diff
@@ -714,6 +714,13 @@ def _gen_new_index(repodata, subdir):
         if any(dep.startswith("mpich >=3.4.1,") for dep in deps):
             _pin_looser(fn, record, "mpich", upper_bound="4.0")
 
+        # TBB 2021 (oneTBB 2021) is incompatible with previous releases.
+        if has_dep(record, "tbb") and record.get('timestamp', 0) < 1614261110000:
+            for i, dep in enumerate(deps):
+                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2017, 2020)) or dep.startswith("tbb >=4.4"):
```
isuruf, line 720:
```suggestion
                if dep == "tbb" or any(dep.startswith(f"tbb >={i}") for i in range(2017, 2021)) or dep.startswith("tbb >=4.4"):
```

traversaro:
I have some doubts on this PR. Gazebo ( see https://github.com/conda-forge/gazebo-feedstock ) is not compatible with 2021.01, so for the immediate future we will continue to pin tbb to 2020.3. However, the repodata patch only covers build that happened before a given point in time (unless I am missing something). However, also recently and in the future (for feedstock that pin tbb) the patch is necessary. While for the future we can also update feedstock that require an old tbb to export the correct constraint, it seems that there was a time period of a few days in in which tbb 2020.03 was the one used by default for building packages, but for which build are not covered by this repodata patch.

For example, the gazebo build `gazebo-11.3.0-h5fc56bc_8` that was built today (so before tbb 2021.1 was released), does not have the required incompatibility with tbb 2021.01 : 
~~~
    "gazebo-11.3.0-h5fc56bc_8.tar.bz2": {
      "build": "h5fc56bc_8",
      "build_number": 8,
      "depends": [
        "boost-cpp >=1.74.0,<1.74.1.0a0",
        "curl >=7.71.1,<8.0a0",
        "ffmpeg >=4.3.1,<5.0a0",
        "freeimage >=3.18.0,<3.19.0a0",
        "libccd >=2.1,<2.2.0a0",
        "libcurl >=7.71.1,<8.0a0",
        "libgcc-ng >=9.3.0",
        "libglu",
        "libignition-cmake2 >=2.6.2,<3.0a0",
        "libignition-common3 >=3.10.1,<4.0a0",
        "libignition-fuel-tools4 >=4.3.0,<5.0a0",
        "libignition-math6 >=6.7.0,<7.0a0",
        "libignition-transport8 >=8.1.0,<9.0a0",
        "libprotobuf >=3.15.3,<3.16.0a0",
        "libsdformat >=9.3.0,<10.0a0",
        "libstdcxx-ng >=9.3.0",
        "libtar >=1.2.20,<1.3.0a0",
        "libuuid >=2.32.1,<3.0a0",
        "libxcb",
        "ogre >=1.10.12,<1.11.0a0",
        "qt >=5.12.9,<5.13.0a0",
        "qwt >=6.1.6,<6.2.0a0",
        "tbb >=2020.2",
        "tbb-devel",
        "tinyxml",
        "tinyxml2 >=8.0.0,<8.1.0a0",
        "xorg-libice",
        "xorg-libsm",
        "xorg-libx11",
        "xorg-libxau",
        "xorg-libxdmcp",
        "xorg-libxext",
        "xorg-libxfixes",
        "xorg-libxrender"
      ],
      "license": "Apache-2.0",
      "md5": "9e0d6e06f2027e7e7edf870f95d06315",
      "name": "gazebo",
      "sha256": "38c872d5f90ffac0ea5cba765cf020d7a2b1a79ac3b5b3109865909fa262edb5",
      "size": 65902153,
      "subdir": "linux-64",
      "timestamp": 1614777077979,
      "version": "11.3.0"
    },
~~~

Do you have any insight on this @AlexVeprev ?

traversaro:
Other affected builds (apparently):
* occt-7.5.1-h8d6c3e3_1.tar.bz2 
* openturns-1.16-py37h2d07b8a_4
* openturns-1.16-py39h2b8f7f0_4
* root-binaries-6.22.6-py37ha638d12_1
* root-binaries-6.22.6-py38hd0710c8_1.tar.bz2
* root-dependencies-6.22.6-py36h9c91e4e_1.tar.bz2

And probably a few others.

isuruf:
the issue was that https://github.com/conda-forge/tbb-feedstock/pull/79 didn't bump the build number.

AlexVeprev:
> didn't bump the build number.

@isuruf but conda-forge/tbb-feedstock#79 updates build number from 2 to 3, do I miss something?

isuruf:
@AlexVeprev, it was already 3. It's a github bug in UI.

AlexVeprev:
@isuruf, hmm, yes, seems I missed fork syncing with the upstream before PR creation, so I updated 2 to 3 in my fork while 3 was already in the original repo... How should I properly resolve it now? Should I create a branch 2020.2 to bump build number there and update timestamp in repodata patch?

isuruf:
> Should I create a branch 2020.2 to bump build number there and update timestamp in repodata patch?

Yes, that's exactly what is needed.

AlexVeprev:
@isuruf, there is the first part: https://github.com/conda-forge/tbb-feedstock/pull/80, it's very simple, but I'd appreciate if you can double check.

AlexVeprev:
The second part is in #127, I've updated a timestamp to Mar 03 2021 22:10:00 UTC (just after deployment of rebuilt 2020.2).

traversaro:
Thanks a lot @AlexVeprev !!

PR #128, opened 2021-03-09
Pin libthrift 0.14 stricter

xhochy:
With 0.14 they added the patch level to the SONAME, fixed in feedstock: https://github.com/conda-forge/thrift-cpp-feedstock/pull/56

Diff: https://gist.github.com/xhochy/be9405c016c9db7687ac1b862c298be4


xhochy:
@conda-forge/core This is ready for a merge.

kkraus14 (approved):

PR #194, opened 2021-12-02
fix tsnecuda

isuruf:
cc @h-vetinari, @erykoff, @beckermr 

This was causing failures on unrelated packages as well because the repodata loading broke.


h-vetinari:
Oh boy, sorry for the mess, and thanks for fixing it... 😑

beckermr:
thank you @isuruf !

PR #85, opened 2020-09-07
Loosen libunwind pin

xhochy:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #199, opened 2021-12-09
Tighten pin on libunwind

zklaus:
Checklist

In conda-forge/libunwind-feedstock#19 we decided to tighten the pin on libunwind from 'x' to 'x.x'. This PR introduces the corresponding repodata patch.

cc @jakirkham

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -851,7 +851,6 @@ def _gen_new_index(repodata, subdir):
             _replace_pin('vector-classes >=1.4.1,<1.5.0a0', 'vector-classes >=1.4.1,<1.4.2a0', deps, record)
 
         _pin_stricter(fn, record, 'libunwind', 'x.x')
-        _replace_pin('libunwind >=1.2.1,<1.3.0a0', 'libunwind >=1.2.1,<2.0.0a0', deps, record)
         _replace_pin('snappy >=1.1.7,<1.1.8.0a0', 'snappy >=1.1.7,<2.0.0.0a0', deps, record)
         _replace_pin('ncurses >=6.1,<6.2.0a0', 'ncurses >=6.1,<6.3.0a0', deps, record)
         _replace_pin('abseil-cpp', 'abseil-cpp ==20190808.*', deps, record)
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -850,7 +850,8 @@ def _gen_new_index(repodata, subdir):
             # ROOT requires vector-classes to be the exact same version as the one used for the build
             _replace_pin('vector-classes >=1.4.1,<1.5.0a0', 'vector-classes >=1.4.1,<1.4.2a0', deps, record)
 
-        _pin_stricter(fn, record, 'libunwind', 'x.x')
+        if has_dep(record, "libunwind"):
+            _pin_stricter(fn, record, 'libunwind', 'x.x')
         _replace_pin('snappy >=1.1.7,<1.1.8.0a0', 'snappy >=1.1.7,<2.0.0.0a0', deps, record)
         _replace_pin('ncurses >=6.1,<6.2.0a0', 'ncurses >=6.1,<6.3.0a0', deps, record)
         _replace_pin('abseil-cpp', 'abseil-cpp ==20190808.*', deps, record)
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -850,6 +850,7 @@ def _gen_new_index(repodata, subdir):
             # ROOT requires vector-classes to be the exact same version as the one used for the build
             _replace_pin('vector-classes >=1.4.1,<1.5.0a0', 'vector-classes >=1.4.1,<1.4.2a0', deps, record)
 
+        _replace_pin('libunwind >=1.5.0,<2.0a0', 'libunwind >=1.5.0,<1.6a0', deps, record)
         _replace_pin('libunwind >=1.2.1,<1.3.0a0', 'libunwind >=1.2.1,<2.0.0a0', deps, record)
         _replace_pin('snappy >=1.1.7,<1.1.8.0a0', 'snappy >=1.1.7,<2.0.0.0a0', deps, record)
         _replace_pin('ncurses >=6.1,<6.2.0a0', 'ncurses >=6.1,<6.3.0a0', deps, record)
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -850,7 +850,7 @@ def _gen_new_index(repodata, subdir):
             # ROOT requires vector-classes to be the exact same version as the one used for the build
             _replace_pin('vector-classes >=1.4.1,<1.5.0a0', 'vector-classes >=1.4.1,<1.4.2a0', deps, record)
 
-        _replace_pin('libunwind >=1.5.0,<2.0a0', 'libunwind >=1.5.0,<1.6a0', deps, record)
+        _pin_stricter(fn, record, 'libunwind', 'x.x')
         _replace_pin('libunwind >=1.2.1,<1.3.0a0', 'libunwind >=1.2.1,<2.0.0a0', deps, record)
         _replace_pin('snappy >=1.1.7,<1.1.8.0a0', 'snappy >=1.1.7,<2.0.0.0a0', deps, record)
         _replace_pin('ncurses >=6.1,<6.2.0a0', 'ncurses >=6.1,<6.3.0a0', deps, record)


zklaus:
@jakirkham, is that about what you had in mind? I must say I feel a little out of my depth here, but happy to learn.

zklaus:

recipe/gen_patch_json.py
```diff
@@ -850,6 +850,7 @@ def _gen_new_index(repodata, subdir):
             # ROOT requires vector-classes to be the exact same version as the one used for the build
             _replace_pin('vector-classes >=1.4.1,<1.5.0a0', 'vector-classes >=1.4.1,<1.4.2a0', deps, record)
 
+        _pin_stricter(fn, record, 'libunwind', 'x.x')
         _replace_pin('libunwind >=1.2.1,<1.3.0a0', 'libunwind >=1.2.1,<2.0.0a0', deps, record)
```
mkitti, line 854: Wouldn't we want to pin `libunwind<1.6` here as well, or am I misunderstanding the task?

jakirkham: Yeah we probably want to do something else here. It's not clear to me atm what we want to do yet though

jakirkham:
Thanks @zklaus! 😄

Yes this is a good start. Had a suggestion above

@mkitti also flagged a line we might need to take a closer look at

zklaus:
Let's take a step back from "How can we patch this?" to look at "What do we want to patch?".

I used [`show_diff.py`](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/master/recipe/show_diff.py) as described in [`readme.md`](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/master/recipe/readme.md) to check which pins would be affected by the `_pin_stricter` solution. I believe those are a superset of the pins we want to change, i.e. all the pins we might want to change are here, but there might be some here that we don't want to change.

Below, you will find a table with all the packages and their respective pin changes, but really, only two changes occur:
- `libunwind >=1.4.0,<2.0a0` gets changed to `libunwind >=1.4.0,<1.5.0a0`
- `libunwind >=1.5.0,<2.0a0` gets changed to `libunwind >=1.5.0,<1.6.0a0`.

Now, the question is whether these are the right changes.
- Do you agree with these changes?
- Do you think these should be changed differently?
- Do you think there are other pins that should be taken into account?

Finally, here is the promised table. Thanks for helping figure this out!

| Package | old pin | new pin |
|-|-|-|
| linux-64::heaptrack-1.2.0-h1b472f0_0.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::heaptrack-1.2.0-h1b472f0_1.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::heaptrack-1.2.0-h26731d4_1.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::julia-1.6.1-h96f571d_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.2-h438a1c7_3.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.2-h96f571d_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.2-hbb0e23a_1.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.3-h2a39549_1.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.3-h438a1c7_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.4-h2a39549_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.4-h34dcedf_1.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::julia-1.6.4-h34dcedf_2.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::py-spy-0.3.10-h2547a8a_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::py-spy-0.3.10-h2547a8a_1.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::py-spy-0.3.11-h2547a8a_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::py-spy-0.3.3-h17340ac_2.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::py-spy-0.3.3-ha1d6e82_1.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::py-spy-0.3.4-h17340ac_0.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::py-spy-0.3.5-h17340ac_0.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::py-spy-0.3.6-h17340ac_0.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::py-spy-0.3.7-h17340ac_0.tar.bz2 | libunwind >=1.4.0,<2.0a0 | libunwind >=1.4.0,<1.5.0a0 |
| linux-64::py-spy-0.3.8-h2547a8a_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-64::py-spy-0.3.9-h2547a8a_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-aarch64::py-spy-0.3.10-h48fc1f3_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-aarch64::py-spy-0.3.10-h48fc1f3_1.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-aarch64::py-spy-0.3.11-h48fc1f3_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-aarch64::py-spy-0.3.8-h48fc1f3_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-aarch64::py-spy-0.3.9-h48fc1f3_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-ppc64le::py-spy-0.3.10-h9b295ce_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-ppc64le::py-spy-0.3.10-h9b295ce_1.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-ppc64le::py-spy-0.3.11-h9b295ce_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-ppc64le::py-spy-0.3.7-he1f5e9e_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |
| linux-ppc64le::py-spy-0.3.8-h9b295ce_0.tar.bz2 | libunwind >=1.5.0,<2.0a0 | libunwind >=1.5.0,<1.6.0a0 |

jakirkham:
Thank for including that detail Klaus! 😄 That looks good to me.

Right the filtering is more about making sure we are not trying to patch package that don't need patching. So a simple `if has_dep(record, "libunwind")` should do the trick.

As to @mkitti's point above we have some old code here that is relaxing a `libunwind` dependency to only depend on major version in some cases (see PR https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/85 for history cc @xhochy). That seems to be the opposite of what we are doing here and so likely needs to be dropped, but will let others comment on that.

recipe/gen_patch_json.py
zklaus: Thanks, @mkitti, good catch. I removed the line and confirmed that the effect of that is that the pinning goes back to 1.3.0a0.

zklaus:
Thanks @jakirkham. As for the filtering, I have nothing against it, but note that [`_pin_stricter`](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/1526b74907359502c6e4c79c0c9989d97ae02652/recipe/gen_patch_json.py#L1491) has this check integrated, so I think it would be a bit redundant.

jakirkham:
cc @conda-forge/core (in case others have thoughts here)

zklaus:
Perhaps I should explain myself a bit better.
The first thing `_pin_stricter` does is build a list of indices into the `depends` list where `fix_dep` (here `libunwind`) appears (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/1526b74907359502c6e4c79c0c9989d97ae02652/recipe/gen_patch_json.py#L1493)). The rest of the function traverses the list, so if it's empty, nothing happens.
The list comprehension that builds the list is equivalent to the generator comprehension used in `has_dep` (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/1526b74907359502c6e4c79c0c9989d97ae02652/recipe/gen_patch_json.py#L335)), hence I think it is ok to only do `_pin_stricter`.

jakirkham:
The idea of this kind of filtering is to not spend time processing things that don't need it. FWIW the rest of the calls to `_pin_stricter` do some conditional filtering first ([for example]( https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/6474712dbf3217010c1d785c5bf7a14b30d44a6b/recipe/gen_patch_json.py#L734-L735 )). So think it is reasonable for us to do something similar here. Is this something you would be able to do @zklaus? 🙂 I think after that we can merge.

zklaus:
Done. Thanks for helping out @jakirkham.

jakirkham:
Thanks Klaus! Also thanks Mark for the review 😄

isuruf:
This is wrong. The previous patch made sure the packages built with 1.2, 1.3, 1.4, 1.5 were installable at the same time. With this patch, this is not possible. The correct patch is to make sure there's a `<1.6` upper bound for packages built with `1.2, 1.3, 1.4, 1.5`.

mkitti:
> The correct patch is to make sure there's a `<1.6` upper bound for packages built with `1.2, 1.3, 1.4, 1.5`

I concur. We need a <1.6 upper bound since Julia with libunwind 1.6.x is not working. How do we do that?

Do you mean this should be done on a per-package basis?
https://github.com/conda-forge/julia-feedstock/pull/136

isuruf:
See https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/203

PR #204, opened 2021-12-21
Add setuptools ceiling

jsignell:
Checklist
* [x] Bumped the build number (if the version is unchanged)

Closes https://github.com/conda-forge/conda-forge.github.io/issues/1575

I'm not sure what the date should be, so I set it to today - ping @isuruf

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -994,6 +994,11 @@ def _gen_new_index(repodata, subdir):
             new_depends = record.get("depends", [])
             new_depends.append("libgcc-ng <=9.3.0")
             record["depends"] = new_depends
+        
+        if record.get("timestamp", 0) < 1640101398654:  # 2021-12-21
+            new_depends = record.get("depends", [])
+            new_depends.append("setuptools <60.0.0")
+            record["depends"] = new_depends
 
         # old CDTs with the conda_cos6 or conda_cos7 name in the sysroot need to
         # conflict with the new CDT and compiler packages
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -999,11 +999,15 @@ def _gen_new_index(repodata, subdir):
         # since packages don't tend to pin setuptools, this raises warnings in old versions
         # https://github.com/conda-forge/conda-forge.github.io/issues/1575
         if (
-            record_name in ["pandas", "distributed", "dask"]
+            record_name in ["pandas", "distributed", "dask-core"]
             and record.get("timestamp", 0) < 1640101398654  # 2021-12-21
         ):
             new_depends = record.get("depends", [])
-            new_depends.append("setuptools <60.0.0")
+            if "setuptools" in new_depends:
+                i = new_depends.index("setuptools")
+                new_depends[i] = "setuptools <60.0.0"
+            else:
+                new_depends.append("setuptools <60.0.0")
             record["depends"] = new_depends
 
         # old CDTs with the conda_cos6 or conda_cos7 name in the sysroot need to
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -995,7 +995,7 @@ def _gen_new_index(repodata, subdir):
             new_depends.append("libgcc-ng <=9.3.0")
             record["depends"] = new_depends
         
-        # setutools started raising a warning when using `LooseVersion` from distutils
+        # setuptools started raising a warning when using `LooseVersion` from distutils
         # since packages don't tend to pin setuptools, this raises warnings in old versions
         # https://github.com/conda-forge/conda-forge.github.io/issues/1575
         if (
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -995,7 +995,13 @@ def _gen_new_index(repodata, subdir):
             new_depends.append("libgcc-ng <=9.3.0")
             record["depends"] = new_depends
         
-        if record.get("timestamp", 0) < 1640101398654:  # 2021-12-21
+        # setutools started raising a warning when using `LooseVersion` from distutils
+        # since packages don't tend to pin setuptools, this raises warnings in old versions
+        # https://github.com/conda-forge/conda-forge.github.io/issues/1575
+        if (
+            record_name in ["pandas", "distributed", "dask"]
+            and record.get("timestamp", 0) < 1640101398654  # 2021-12-21
+        ):
             new_depends = record.get("depends", [])
             new_depends.append("setuptools <60.0.0")
             record["depends"] = new_depends


jsignell:

beckermr:
cc @conda-forge/core

recipe/gen_patch_json.py
```diff
@@ -994,6 +994,11 @@ def _gen_new_index(repodata, subdir):
             new_depends = record.get("depends", [])
             new_depends.append("libgcc-ng <=9.3.0")
             record["depends"] = new_depends
+        
+        if record.get("timestamp", 0) < 1640101398654:  # 2021-12-21
```
isuruf, line 998: We should restrict this to pandas. We don't want to add a setuptools dependency to *all* packages.
```suggestion
        if record_name == "pandas" and record.get("timestamp", 0) < 1640101398654:  # 2021-12-21
```

jsignell: Oh I don't think this only applies to pandas though. This applies to every package that ever used `LooseVersion` from `distutils` without pinning setuptools. Right?

jsignell: But maybe pandas is a good enough place to start I suppose.

beckermr: If we can put together a census of packages, we can add them here later. 

Also, @CJ-Wright may have some tools to help us examine which packages with which versions use this symbol.

jsignell: Ah ok yeah I was going to start tracing back which versions different packages switch to packaging. I don't think it's too terrible to do this based on timestamp though if that ends up being hard. 

Ping @@jorisvandenbossche to make sure pandas is ok with this patch.

jakirkham:
cc @conda-forge/pandas

jsignell:
Should I remove pandas from this PR and just do dask?

beckermr:
yes. let's do that for now. We can add others as needed.

recipe/gen_patch_json.py
```diff
@@ -994,6 +994,17 @@ def _gen_new_index(repodata, subdir):
             new_depends = record.get("depends", [])
             new_depends.append("libgcc-ng <=9.3.0")
             record["depends"] = new_depends
+        
+        # setutools started raising a warning when using `LooseVersion` from distutils
```
jakirkham, line 998:
```suggestion
        # setuptools started raising a warning when using `LooseVersion` from distutils
```

recipe/gen_patch_json.py
```diff
@@ -994,6 +994,17 @@ def _gen_new_index(repodata, subdir):
...
+        # since packages don't tend to pin setuptools, this raises warnings in old versions
+        # https://github.com/conda-forge/conda-forge.github.io/issues/1575
+        if (
+            record_name in ["pandas", "distributed", "dask"]
```
jakirkham, line 1002: Looks like `dask` doesn't list this as a dependency. Only `distributed` does

```suggestion
            record_name == "distributed"
```

recipe/gen_patch_json.py
```diff
@@ -994,6 +994,17 @@ def _gen_new_index(repodata, subdir):
...
+            and record.get("timestamp", 0) < 1640101398654  # 2021-12-21
+        ):
+            new_depends = record.get("depends", [])
+            new_depends.append("setuptools <60.0.0")
```
jakirkham, line 1006: @jrbourbeau do we already have a `distributed` feedstock issue tracking adding this constraint?

jakirkham (commented):
Thank Julia! 😄

Had a couple suggestions below 🙂

recipe/gen_patch_json.py
jsignell: The python dependency from conda-forge pulls in setuptools implicitly even if it's not listed. So I think we should actually be adding this dependency to dask-core.

```suggestion
            record_name in ["dask-core", "distributed"]
```

recipe/gen_patch_json.py
```diff
@@ -994,6 +994,17 @@ def _gen_new_index(repodata, subdir):
...
             record["depends"] = new_depends
+        
+        # setutools started raising a warning when using `LooseVersion` from distutils
+        # since packages don't tend to pin setuptools, this raises warnings in old versions
+        # https://github.com/conda-forge/conda-forge.github.io/issues/1575
+        if (
+            record_name in ["pandas", "distributed", "dask"]
+            and record.get("timestamp", 0) < 1640101398654  # 2021-12-21
+        ):
+            new_depends = record.get("depends", [])
```
jsignell, line 1005: I had not realized that the distributed feedstock does explicitly include setuptools. I think this change should handle that:

```suggestion
            new_depends = record.get("depends", [])
            if "setuptools" in new_depends:
                new_depends.remove("setuptools")
```

recipe/gen_patch_json.py
jsignell: I just made one https://github.com/conda-forge/distributed-feedstock/issues/192

recipe/gen_patch_json.py
beckermr: Don't do this. People have run deps on setuptools for lots of reasons.

beckermr: ohhhh nvm you are removing the old one.

recipe/gen_patch_json.py
jakirkham: `setuptools` is [not a `python` dependency]( https://github.com/conda-forge/python-feedstock/blob/1bb50920fe462e1a30ebda8c3fdabb30fe4bde0c/recipe/meta.yaml#L183-L185 ).

That said, [`conda` can pull in `pip` when installing `python`]( https://docs.conda.io/projects/conda/en/latest/user-guide/configuration/use-condarc.html#add-pip-as-python-dependency-add-pip-as-python-dependency ) and [`pip` does pull in `setuptools`]( https://github.com/conda-forge/pip-feedstock/blob/774427c945373fa7ff1386019c57635301a35019/recipe/meta.yaml#L25 ). So guessing this is what you referring to.

Looking at Dask itself, it doesn't appear to use [`pkg_resources`]( https://github.com/dask/dask/search?q=pkg_resources ) or [`distutils`]( https://github.com/dask/dask/search?q=distutils ). The only [uses of `setuptools` are in `setup.py` and `versioneer`]( https://github.com/dask/dask/search?p=1&q=setuptools ). So it seems that if we were to add a constraint, we should add a `setuptools` constraint to `requirements/host` in `dask-core`.

recipe/gen_patch_json.py
jakirkham: Normally what we do is find the constraint and replace it

jorisvandenbossche (commented):
I think it's fine to keep pandas, the logic sounds correct. 

One thing though: shouldn't we put a max version check on the packages, so we don't apply this patch on newer versions? At least for pandas, the usage of distutils is already fixed, and this is already released starting from pandas 1.3.0

beckermr:
> One thing though: shouldn't we put a max version check on the packages, so we don't apply this patch on newer versions? At least for pandas, the usage of distutils is already fixed, and this is already released starting from pandas 1.3.0

Yes we should, but use the package timestamps. Those are easier to manage, especially for builds of the same version where some need patches and others do not.

jsignell:
I already have a timestamp constraint on this don't I?

beckermr:
Ahh good. Then we shouldn't need the max version check since the timestamp limits to stuff shipped last year.

recipe/gen_patch_json.py
jakirkham: @jrbourbeau, do you have any thoughts here?

jsignell: Ok I changed this to do dask-core rather than dask.

recipe/gen_patch_json.py
jsignell: I am now finding the constraint and replacing as suggested.

jakirkham (approved):

jakirkham:
@beckermr any more thoughts here? 🙂

jakirkham:
Thanks all! 😄

rgommers:
Just seeing this now (and learned something, thanks!) - can we please add `numpy` and `scipy`? We do not support setuptools >= 60.0, and have more problems than just `LooseVersion` usage. There are pins in our `main` branches now, and I don't expect to remove them any time soon.

beckermr:
Yeah, that sounds good! Go ahead and make a PR!

isuruf:
Note that `setuptools<60` is unsupported in setuptools-feedstock right now. If you want to see <60, please step up.

rgommers:
@isurug what does unsupported mean here? `setuptools` is a pure Python package, and `59.x` works fine - there shouldn't be anything to do.

isuruf:
Even though it is a pure python package, it's not a noarch: python package, so it needs a rebuild when for example python 3.11 or pypy3.8 comes along. We carry patches in setuptools to fix conda and they depend on python version, so they have to be refreshed when a new python version comes along. So, we need to have a lts branch on setuptools for `59.x` which we don't have right now.

rgommers:
Okay, that makes sense. The patches in https://github.com/conda-forge/setuptools-feedstock/blob/master/recipe/meta.yaml are fairly minimal so rebuilds should work without changes I'd expect. I'd prefer helping out with that rather than debugging issues due to setuptools versions that are known to be problematic and are untested in the numpy/scipy `main` branches.

The other alternative is setting `SETUPTOOLS_USE_DISTUTILS=stdlib` in recipes, but that's also a completely untested config.

Is there anything special to do for an LTS branch? I don't see it mentioned in the conda-forge docs. Happy to make a PR if that is better to do now, if you can point me to an example.

isuruf:
@ocefpaf, do we have any docs on LTS branches?

beckermr:
These are called Abi branches or similar in the docs. Look in the bot sections our outreachy intern wrote

beckermr:
See here https://conda-forge.org/docs/maintainer/conda_forge_yml.html#bot

Add a branch to the feedstock with ci skip in the commit and the put that branch in the cf yaml on the master branch. Then you can merge v59 or w/e into that branch. The bot will make Abi migration prs to that branch as needed.

rgommers:
Thanks. I'll see if I can tweak the docs a bit to mention LTS, so it's easier to search for.

beckermr:
Yeah we have a maintain multiple versions section. This info could go there. Thanks for all of the help!

zzhengnan:
Hi all. I didn't know a patch like this was even possible. Thanks for the discussion so far!

Having reading through the discussions in this PR, https://github.com/conda-forge/conda-forge.github.io/issues/1575, and https://github.com/dask/dask/pull/8509, I'm not sure I fully understand the rationale behind this patch. I could be missing something obvious, but is there a reason _other than_ to spare projects like Dask -- which choose to interpret warnings as errors during their CI process -- the hassle of having to manually add the `setuptools<60.0.0` pin in their CI environments?

My colleagues and I recently ran into an issue that we believe was related to this patch. First, some context
- Our application is written in Python 3.7 and heavily relies on the PyData stack
- We generate a "frozen" environment file (via `conda export`) at the end of our testing process
- The same file is subsequently used to replicate the testing environment for our end users

Our latest frozen environment was generated on 1/4/2022 and looks as follows. This date happened to fall _after_ `setuptools` 60.x.x was released (on [12/21/2021](https://anaconda.org/conda-forge/setuptools/files?version=60.0.3)) but _before_ the `<60.0.0` patch was merged (on 1/14/2021), which is why 60.2.0 was picked up as opposed to 59.x.x.

```yaml
dependencies:
  - dask-core=2.20.0
  - distributed=2.30.1
  - pandas=1.0.5
  - setuptools=60.2.0
  # ... many others
```

The problem is that now when we attempt to run `conda env create` based off this "frozen" environment, conda complains about version conflicts since the repodata has been patched to require `setuptools<60.0.0`.

Would it be reasonable to assume that any project that has frozen their environment between 12/21/2021 and 1/14/2021 will run into this exact issue?

zzhengnan:
Thanks for the suggestion, @isuruf.

If I generate a lock file with `conda list --explicit` and subsequently attempt to recreate the environment, is it _guaranteed_ that the result will be a _consistent_ environment.

I've had issues with `conda list --explicit` in the past where the recreated environment is not consistent, causing conda to fail whenever I attempt to install any additional packages into the new environment.

PR #212, opened 2022-01-15
Add a `setuptools<60.0` dependency for numpy and scipy

rgommers:
See gh-204 for discussion. For SciPy 1.9.0 the pin may be removed if we build the conda-forge package with Meson rather than `numpy.distutils` (as is currently planned).

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -994,13 +994,18 @@ def _gen_new_index(repodata, subdir):
             new_depends = record.get("depends", [])
             new_depends.append("libgcc-ng <=9.3.0")
             record["depends"] = new_depends
-        
+
         # setuptools started raising a warning when using `LooseVersion` from distutils
         # since packages don't tend to pin setuptools, this raises warnings in old versions
         # https://github.com/conda-forge/conda-forge.github.io/issues/1575
+        # This is what the time-based pin is for.
+        # NumPy and SciPy do not support setuptools >= 60.0, and perhaps never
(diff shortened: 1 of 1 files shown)


recipe/gen_patch_json.py
```diff
@@ -994,13 +994,18 @@ def _gen_new_index(repodata, subdir):
...
             new_depends.append("libgcc-ng <=9.3.0")
             record["depends"] = new_depends
-        
+
         # setuptools started raising a warning when using `LooseVersion` from distutils
         # since packages don't tend to pin setuptools, this raises warnings in old versions
         # https://github.com/conda-forge/conda-forge.github.io/issues/1575
+        # This is what the time-based pin is for.
+        # NumPy and SciPy do not support setuptools >= 60.0, and perhaps never
+        # will, so they get a permanent pin. Other packages using
```
isuruf, line 1003: Permanent pins should go in the recipe, instead of repodata patching.

rgommers: Okay thanks. That wasn't clear to me from the discussion in gh-204. Closing, will resubmit on the individual feedstocks.

isuruf: Sorry, I should have made it clear. We should patch the older packages here with a timestamp check.

rgommers:
On second thought, this patching is unnecessary since it's for runtime dependencies not build dependencies. This was not clear to me from the discussion on gh-204 (it should have been from reading the code though). Sorry for the noise.

PR #220, opened 2022-02-02
patch holoview <=1.14.7 to <py310

ocefpaf:
It has been a while since the last time I sent a PR like this one. I appreciate someone taking a look if I got everything right.

More context in https://discourse.holoviz.org/t/datashading-is-not-available-error-with-python-3-10-2/3333

The next, still unreleased, version of holoviews fixed this issue and we should be OK limiting the patch to versions 1.14.7 and below.

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -1290,6 +1290,13 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
             elif record["version"] == "0.7.14":
                 _replace_pin("python >=2.7", "python >=2.7,<3.10", deps, record)
 
+        # Retroactively pin hloviews to Python < 3.10 b/c they import Callable directly and that was removed in py310.
+        # This is already fixed in next verision of holoviews >1.14.7 (unrelease).
+        if record_name == "holoviews":
+            deps = record["depends"]
+            if  record['version'] <= "1.14.7":
+                _replace_pin("python", "python >=2.7,<3.10", deps, record)
+
         if record_name == "tsnecuda":
             # These have dependencies like
             # - libfaiss * *_cuda


recipe/gen_patch_json.py
```diff
@@ -1290,6 +1290,13 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
             elif record["version"] == "0.7.14":
                 _replace_pin("python >=2.7", "python >=2.7,<3.10", deps, record)
 
+        # Retroactively pin hloviews to Python < 3.10 b/c they import Callable directly and that was removed in py310.
+        # This is already fixed in next verision of holoviews >1.14.7 (unrelease).
+        if record_name == "holoviews":
+            deps = record["depends"]
+            if  record['version'] <= "1.14.7":
```
beckermr, line 1297: This is a straight string comparison. I think unless I am missing some magic on encodings that we'll want to cast to a version object or something before doing this.

ocefpaf: Yeah, I was going to do a == first and for that it would have been ok. After some investigation I decided to do all previous version. I'll do a version object. Any suggestions? I know there a few options available but I never used any of them.

beckermr: I'm pretty sure if you search this source file, there is one already imported.

ocefpaf: @isuruf, any suggestions? It would be nice to make it consistent. I see many different ways to do it.

xylar: It doesn't look like @isuruf had a chance to respond.  It looks like this approach would work:
https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/d94374e7e54d79a06b881fcfbce4235b5c3c41d5/recipe/gen_patch_json.py#L551-L553
So in your case:
``` python
            pversion = pkg_resources.parse_version(record['version'])
            pv_1147 = pkg_resources.parse_version('1.14.7')
            if pversion <= pv_1147:
```

ocefpaf: I forgot about this one. While it would be nive to standardize this in our script. This particular package doesn't really need a repodata patch. It is controversial but one can argue that 99% of the package works with Python 3.10. Just one small functionality that does not. Also, the latest version is 100% compatible, so the patch would break workflow of folks who are pinning this at the moment.

PR #229, opened 2022-02-23
patch wrong run_export for ffmpeg 4.3

h-vetinari:
Following the steps outlined [here](https://github.com/conda-forge/ffmpeg-feedstock/pull/115#issuecomment-1020619231).

Timestamp will need adjusting until after the builds of https://github.com/conda-forge/ffmpeg-feedstock/pull/119 go live.

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -785,7 +785,9 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if any(dep == "libflang" or dep.startswith("libflang >=5.0.0") for dep in deps) and record.get('timestamp', 0) < 1611789153000:
             record["depends"].append("libflang <6.0.0.a0")
 
-        if any(dep == "ffmpeg" for dep in deps) and record.get('timestamp', 0) < 1645603462892:
+        # fix run_export from packages built against 4.3; it's corrected now, but the solver
+        # may potentially still pick up an old ffmpeg-build as build dep for something else
+        if any(dep.startswith(f"ffmpeg >=4.3.{p},<5.0a0") for dep in deps for p in [0, 1, 2]):
             # https://github.com/conda-forge/ffmpeg-feedstock/pull/115#issuecomment-1020619231
             record["depends"].append("ffmpeg <4.4")
 
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -787,7 +787,7 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
 
         # fix run_export from packages built against 4.3; it's corrected now, but the solver
         # may potentially still pick up an old ffmpeg-build as build dep for something else
-        if any(dep == f"ffmpeg >=4.3.{p},<5.0a0" for dep in deps for p in [0, 1, 2]):
+        if any(dep == f"ffmpeg >=4.3.{p},<5.0a0" for dep in deps for p in [0, 1, 2]) and record.get('timestamp', 0) < 1645651093167:
             # https://github.com/conda-forge/ffmpeg-feedstock/pull/115#issuecomment-1020619231
             _pin_stricter(fn, record, "ffmpeg", "x.x")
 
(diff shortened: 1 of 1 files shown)


h-vetinari:
Output of `python show_diff.py`:

h-vetinari:
Aside from review, waiting for https://github.com/conda-forge/ffmpeg-feedstock/pull/119 to be merged.

hmaarrfk:
Is it worth limiting this addition to packages built with FFMPEG 4?

hmaarrfk:
There are also some packages that do not pin against the ABI that wouldn't be caught by future migrators.

One that i see is `imageio-ffmpeg` which only depends on the executable:
https://github.com/conda-forge/imageio-ffmpeg-feedstock/blob/master/recipe/meta.yaml#L27

Is there a way to exclude that package from the list?

h-vetinari:
> Is it worth limiting this addition to packages built with FFMPEG 4?

That's IMO not an issue, as it would be compatible (resp no-op) with existing constraints.

> There are also some packages that do not pin against the ABI that wouldn't be caught by future migrators.

That however is a good point. Can we determine from the `record` which dependencies are host? It seems no other patch in `gen_patch_json.py` mentions "host"...

hmaarrfk:
Maybe we can write a rule that is like:

```
if "<5" then make it "<=4.3"; else leave unchanged
```

isuruf:
There might have been packages built with 4.4, so please check if the constraint is `>=4.3.a0,<5.0` and change it.

h-vetinari:
Good idea to go for the run-export directly, which I've done in the last commit. Could use regular expression, but that doesn't seem to be common in `gen_patch_json.py`.

Updated diff:

h-vetinari:
Since the condition is now much sharper (and is applicable only for builds against 4.3), as well as due to @hmaarrfk's [warning](https://github.com/conda-forge/ffmpeg-feedstock/pull/119#issuecomment-1048767284) that old 4.3 builds may still be picked up by the solver, I removed the timeboxing.

recipe/gen_patch_json.py
```diff
@@ -785,6 +785,12 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if any(dep == "libflang" or dep.startswith("libflang >=5.0.0") for dep in deps) and record.get('timestamp', 0) < 1611789153000:
             record["depends"].append("libflang <6.0.0.a0")
 
+        # fix run_export from packages built against 4.3; it's corrected now, but the solver
+        # may potentially still pick up an old ffmpeg-build as build dep for something else
+        if any(dep.startswith(f"ffmpeg >=4.3.{p},<5.0a0") for dep in deps for p in [0, 1, 2]):
```
isuruf, line 790:
```suggestion
        if any(dep == f"ffmpeg >=4.3.{p},<5.0a0" for dep in deps for p in [0, 1, 2]):
```
or 
```suggestion
        if any(dep.startswith("ffmpeg >=4.3) for dep in deps):
```

recipe/gen_patch_json.py
```diff
@@ -785,6 +785,12 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
...
+            # https://github.com/conda-forge/ffmpeg-feedstock/pull/115#issuecomment-1020619231
+            record["depends"].append("ffmpeg <4.4")
```
isuruf, line 792:
```suggestion
            _pin_stricter(fn, record, "ffmpeg", "x.x")
```

isuruf:
Can you add a timestamp too?

h-vetinari:
> Can you add a timestamp too?

I intentionally left that out. Because the `>=4.3.x,<5.0a0` can only come from a 4.3 run_export, and if the solver is picky, it might still resolve to the old builds with the looser run exports. Hence I'd like to keep correcting this also in the future (and I don't think there's a risk that it applies to anything else, e.g. builds against 4.4)

isuruf:
That's not the policy for repodata patches. The patches are applied to only past packages, not new packages.
If you are worried about that, please send a PR to mark the older packages as broken.

h-vetinari:
> That's not the policy for repodata patches. 

I was trying to heed the warning by @hmaarrfk, but it's not something I'm invested in (and I thought that with good enough reason, untimeboxed patches can still happen, e.g. the ones involved in https://github.com/conda-forge/llvmdev-feedstock/issues/141) - commit pushed.

isuruf:
> I thought that with good enough reason, untimeboxed patches can still happen, e.g. the ones involved in https://github.com/conda-forge/llvmdev-feedstock/issues/141

No, we want to get rid of them which is why https://github.com/conda-forge/llvmdev-feedstock/issues/141 was opened.

PR #232, opened 2022-02-25
Add missing __glibc dependency for libgdal

zklaus:
In conda-forge/gdal-feedstock#604 (build number 3 of version 3.4.1), a new sysroot for linux64 builds was introduced for gdal.
Due to confusion from the multi-output nature of the recipe, the correct sysroot dependency was included in the top-level dependencies, but missing from the `libgdal` output. As a consequence, the `libgdal` package is missing the runtime dependency on `__glibc >=2.17`.

The correct dependency was introduced for build numbers >=4 in conda-forge/gdal-feedstock#607, but build number 3 needs a repodata fix to avoid breakage on systems with older glibc.

Closes conda-forge/gdal-feedstock#606

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -1380,6 +1380,12 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
             record["depends"].remove("six")
             _replace_pin("toml", "toml >=0.10.2,<1.0.0", record["depends"], record)
             _replace_pin("xmltodict", "xmltodict >=0.12.0,<1.0.0", record["depends"], record)
+
+        if (record_name == "libgdal"
+            and subdir == "linux-64"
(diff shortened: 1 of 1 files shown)


xylar:
This patch looks correct to me.

hmaarrfk:
Can you verify that the output of the show_diff script provides what you want?

zklaus:
Yes. I did that before opening this PR and it shows exactly what is expected, i.e.
```
linux-64::libgdal-3.4.1-h42fb820_3.tar.bz2
+    "__glibc >=2.17,<3.0.a0",
linux-64::libgdal-3.4.1-hc3d8651_3.tar.bz2
+    "__glibc >=2.17,<3.0.a0",
```
I compared this with the corresponding `info/index.json` from build number 4.

PR #303, opened 2022-08-04
Stricter pin for packages built against antic

saraedum:
On macOS, the C library antic is called libantic-0.2.4.dylib and libantic-0.2.5.dylib. Therefore, we require a strict pin: when built
against libantic-0.2.4.dylib, we cannot use the shared library libantic-0.2.5.dylib even though their ABI would be compatible.

There is a fix upstream to rename the dylib as recommended by Apple, i.e., call it libantic.0.dylib. Until that fix is released, the pin in the antic-feedstock is now x.x.x for macOS. However, for packages built against previous builds of antic, we need to change the pin from x.x to x.x.x.


saraedum:
The fix in the antic feedstock is here: https://github.com/conda-forge/antic-feedstock/pull/10

The upstream issue is here: https://github.com/wbhart/antic/issues/80

PR #309, opened 2022-08-19
Print "Downloading: ..." messages to stderr

maresb:
This makes it possible to run `./show_diff.py > repodata.diff` without polluting the output with `Downloading: ...` messages.

This is useful for large diffs.


maresb:
@conda-forge/core, requesting review for a simple and helpful change (three lines of code)

PR #325, opened 2022-10-02
Patch for `cuda-python` 11.7.1

leofang:
Part of fixing https://github.com/conda-forge/cuda-python-feedstock/issues/15.

I did a quick search and it seems currently only `rmm` and `numba` depend on `cuda-python`.

See the diff here:
<details><summary>CLICK ME</summary>

```diff
Downloading: https://conda.anaconda.org/conda-forge/noarch/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/noarch/repodata.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-64/repodata.json.bz2
linux-64::numba-0.55.2-py310ha5257ce_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.55.2-py37h43839f2_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.55.2-py38hdc3674a_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.55.2-py39h66db6d7_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py310ha5257ce_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py310ha5257ce_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py37hf081915_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py37hf081915_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py38h9a4aae9_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py38h9a4aae9_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py39h61ddf18_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-64::numba-0.56.2-py39h61ddf18_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::rmm-22.06.00-py310h1de6c43_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-64::rmm-22.06.00-py38h81e7143_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-64::rmm-22.06.00-py39hbcfe0b2_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-64::rmm-22.08.00-py310h1de6c43_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-64::rmm-22.08.00-py38h81e7143_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-64::rmm-22.08.00-py39hbcfe0b2_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
Downloading: https://conda.anaconda.org/conda-forge/linux-armv7l/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-armv7l/repodata.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-aarch64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-aarch64/repodata.json.bz2
linux-aarch64::numba-0.55.2-py310h7e061d2_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.55.2-py37hea1f400_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.55.2-py38h90f1751_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.55.2-py39h780101b_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py310h7e061d2_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.56.2-py310h7e061d2_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.56.2-py37h1c995b0_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py37h1c995b0_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.56.2-py38h4c0ec4d_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py38h4c0ec4d_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py39h58e5d4c_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py39h58e5d4c_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::rmm-22.06.00-py310h9691930_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-aarch64::rmm-22.06.00-py38h9c9d46e_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-aarch64::rmm-22.06.00-py39hebf1436_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-aarch64::rmm-22.08.00-py310h9691930_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-aarch64::rmm-22.08.00-py38h9c9d46e_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-aarch64::rmm-22.08.00-py39hebf1436_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
Downloading: https://conda.anaconda.org/conda-forge/linux-ppc64le/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-ppc64le/repodata.json.bz2
linux-ppc64le::numba-0.55.2-py310h782a0c3_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-ppc64le::numba-0.55.2-py37h483c640_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-ppc64le::numba-0.55.2-py38h4c6250a_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-ppc64le::numba-0.55.2-py39ha474065_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-ppc64le::rmm-22.06.00-py310hadc03bd_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-ppc64le::rmm-22.06.00-py38habd7850_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-ppc64le::rmm-22.06.00-py39hb0a7ebe_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-ppc64le::rmm-22.08.00-py310hadc03bd_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-ppc64le::rmm-22.08.00-py38habd7850_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-ppc64le::rmm-22.08.00-py39hb0a7ebe_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
Downloading: https://conda.anaconda.org/conda-forge/osx-64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/osx-64/repodata.json.bz2
osx-64::numba-0.55.2-py310hab1bff6_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.55.2-py37h7687b59_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.55.2-py38h1f551d2_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
osx-64::numba-0.55.2-py39hc37845d_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py310h62db5c2_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py310h62db5c2_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py37h7da6166_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py37h7da6166_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py38hab356c4_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
osx-64::numba-0.56.2-py38hab356c4_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py39hf6b9d82_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py39hf6b9d82_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
Downloading: https://conda.anaconda.org/conda-forge/osx-arm64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/osx-arm64/repodata.json.bz2
osx-arm64::numba-0.55.2-py310ha67b2a8_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.55.2-py38h25e2f74_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.55.2-py39hc0d1af8_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py310h3124f1e_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py310h3124f1e_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py38h947a10e_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py38h947a10e_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
osx-arm64::numba-0.56.2-py39h251cc7c_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py39h251cc7c_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
Downloading: https://conda.anaconda.org/conda-forge/win-32/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/win-32/repodata.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/win-64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/win-64/repodata.json.bz2
win-64::numba-0.55.2-py310h77579ad_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.55.2-py37hc29f945_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.55.2-py38h860efb6_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.55.2-py39hb8cd55e_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py310h19bcfe9_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py310h19bcfe9_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py37h9c05cb2_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
win-64::numba-0.56.2-py37h9c05cb2_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py38h3873db4_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py38h3873db4_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py39h99ae161_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py39h99ae161_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
```

</p>
</details>

cc: @kkraus14 @jakirkham @conda-forge/rmm @conda-forge/numba @conda-forge/cuda-python 

Checklist

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -808,6 +808,22 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if i >= 0:
             deps[i] = "cudatoolkit >=11.2,<12.0a0"
 
+        # cuda-python 11.7.1 is ABI-breaking, currently it impacts rmm & numba
+        if record_name == "rmm":
+            if pkg_resources.parse_version(record["version"]) < pkg_resources.parse_version("22.08"):
+                # these were built with cuda-python <= 11.7.0
+                _pin_stricter(fn, record, "cuda-python", max_pin="x", upper_bound="11.7.1")
+            if (pkg_resources.parse_version(record["version"]) == pkg_resources.parse_version("22.08")
+                and record['build_number'] == 0):
+                # this is built with cuda-python 11.7.1
+                _replace_pin("cuda-python >=11.7,<12.0a0", "cuda-python >=11.7.1,<12.0a0", deps, record)
+        if (record_name == "numba"
+            and pkg_resources.parse_version(record["version"]) <= pkg_resources.parse_version("0.56.2")):
+            for i, dep in enumerate(record.get("constrains", [])):
+                if dep.startswith("cuda-python"):
+                    record["constrains"][i] = "cuda-python >=11.6,<11.7.1"
+                    break
+
         if record.get('timestamp', 0) < 1663795137000:
             if any(dep.startswith("arpack >=3.7") for dep in deps):
                 _pin_looser(fn, record, "arpack", max_pin="x.x")
@@ -2112,7 +2128,7 @@ def _relax_libssh2_1_x_pinning(fn, record):
         depends[dep_idx] = "libssh2 >=1.8.0,<2.0.0a0"
 
 
-cb_pin_regex = re.compile(r"^>=(?P<lower>\d(\.\d+)*a?),<(?P<upper>\d(\.\d+)*)a0$")
+cb_pin_regex = re.compile(r"^>=(?P<lower>\d{1,2}(\.\d+)*a?),<(?P<upper>\d{1,2}(\.\d+)*)a0$")
 
 def _pin_stricter(fn, record, fix_dep, max_pin, upper_bound=None):
     depends = record.get("depends", ())
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -823,11 +823,6 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
                 if dep.startswith("cuda-python"):
                     record["constrains"][i] = "cuda-python >=11.6,<11.7.1"
                     break
-        if (record_name == "cuda-python"
-            and pkg_resources.parse_version(record["version"]) == pkg_resources.parse_version("11.7.1")
-            and record['build_number'] == 0):
-            # add run_exports
-            deps.append("cuda-python >=11.7.1,<12.0a0")
 
         if record.get('timestamp', 0) < 1663795137000:
             if any(dep.startswith("arpack >=3.7") for dep in deps):
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -823,6 +823,11 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
                 if dep.startswith("cuda-python"):
                     record["constrains"][i] = "cuda-python >=11.6,<11.7.1"
                     break
+        if (record_name == "cuda-python"
+            and pkg_resources.parse_version(record["version"]) == pkg_resources.parse_version("11.7.1")
+            and record['build_number'] == 0):
+            # add run_exports
+            deps.append("cuda-python >=11.7.1,<12.0a0")
 
         if record.get('timestamp', 0) < 1663795137000:
             if any(dep.startswith("arpack >=3.7") for dep in deps):
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -808,7 +808,7 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if i >= 0:
             deps[i] = "cudatoolkit >=11.2,<12.0a0"
 
-        # cuda-python 11.7.1 is ABI-breaking, currently it impacts rmm & numba
+        # cuda-python 11.7.1 is ABI-breaking, currently it impacts only rmm
         if record_name == "rmm":
             if pkg_resources.parse_version(record["version"]) < pkg_resources.parse_version("22.08"):
                 # these were built with cuda-python <= 11.7.0
@@ -817,12 +817,6 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
                 and record['build_number'] == 0):
                 # this is built with cuda-python 11.7.1
                 _replace_pin("cuda-python >=11.7,<12.0a0", "cuda-python >=11.7.1,<12.0a0", deps, record)
-        if (record_name == "numba"
-            and pkg_resources.parse_version(record["version"]) <= pkg_resources.parse_version("0.56.2")):
-            for i, dep in enumerate(record.get("constrains", [])):
-                if dep.startswith("cuda-python"):
-                    record["constrains"][i] = "cuda-python >=11.6,<11.7.1"
-                    break
 
         if record.get('timestamp', 0) < 1663795137000:
             if any(dep.startswith("arpack >=3.7") for dep in deps):


leofang:
New diff:

<details><summary>CLICK ME</summary>
<p>

```diff
Downloading: https://conda.anaconda.org/conda-forge/noarch/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/noarch/repodata.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-64/repodata.json.bz2
linux-64::cuda-python-11.7.1-py310h34d4096_0.tar.bz2
-    "python_abi 3.10.* *_cp310"
+    "python_abi 3.10.* *_cp310",
+    "cuda-python >=11.7.1,<12.0a0"
linux-64::cuda-python-11.7.1-py37h9ef1b40_0.tar.bz2
-    "python_abi 3.7.* *_cp37m"
+    "python_abi 3.7.* *_cp37m",
+    "cuda-python >=11.7.1,<12.0a0"
linux-64::cuda-python-11.7.1-py38h7525318_0.tar.bz2
-    "python_abi 3.8.* *_cp38"
+    "python_abi 3.8.* *_cp38",
+    "cuda-python >=11.7.1,<12.0a0"
linux-64::cuda-python-11.7.1-py39h1eff087_0.tar.bz2
-    "python_abi 3.9.* *_cp39"
+    "python_abi 3.9.* *_cp39",
+    "cuda-python >=11.7.1,<12.0a0"
linux-64::numba-0.55.2-py310ha5257ce_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.55.2-py37h43839f2_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.55.2-py38hdc3674a_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.55.2-py39h66db6d7_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py310ha5257ce_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py310ha5257ce_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py37hf081915_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py37hf081915_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py38h9a4aae9_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py38h9a4aae9_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::numba-0.56.2-py39h61ddf18_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-64::numba-0.56.2-py39h61ddf18_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-64::rmm-22.06.00-py310h1de6c43_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-64::rmm-22.06.00-py38h81e7143_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-64::rmm-22.06.00-py39hbcfe0b2_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-64::rmm-22.08.00-py310h1de6c43_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-64::rmm-22.08.00-py38h81e7143_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-64::rmm-22.08.00-py39hbcfe0b2_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
Downloading: https://conda.anaconda.org/conda-forge/linux-armv7l/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-armv7l/repodata.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-aarch64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-aarch64/repodata.json.bz2
linux-aarch64::cuda-python-11.7.1-py310h01c92d2_0.tar.bz2
-    "python_abi 3.10.* *_cp310"
+    "python_abi 3.10.* *_cp310",
+    "cuda-python >=11.7.1,<12.0a0"
linux-aarch64::cuda-python-11.7.1-py37hef93bea_0.tar.bz2
-    "python_abi 3.7.* *_cp37m"
+    "python_abi 3.7.* *_cp37m",
+    "cuda-python >=11.7.1,<12.0a0"
linux-aarch64::cuda-python-11.7.1-py38h18becac_0.tar.bz2
-    "python_abi 3.8.* *_cp38"
+    "python_abi 3.8.* *_cp38",
+    "cuda-python >=11.7.1,<12.0a0"
linux-aarch64::cuda-python-11.7.1-py39hacd4443_0.tar.bz2
-    "python_abi 3.9.* *_cp39"
+    "python_abi 3.9.* *_cp39",
+    "cuda-python >=11.7.1,<12.0a0"
linux-aarch64::numba-0.55.2-py310h7e061d2_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.55.2-py37hea1f400_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.55.2-py38h90f1751_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.55.2-py39h780101b_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py310h7e061d2_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.56.2-py310h7e061d2_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.56.2-py37h1c995b0_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py37h1c995b0_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-aarch64::numba-0.56.2-py38h4c0ec4d_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py38h4c0ec4d_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py39h58e5d4c_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::numba-0.56.2-py39h58e5d4c_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-aarch64::rmm-22.06.00-py310h9691930_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-aarch64::rmm-22.06.00-py38h9c9d46e_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-aarch64::rmm-22.06.00-py39hebf1436_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-aarch64::rmm-22.08.00-py310h9691930_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-aarch64::rmm-22.08.00-py38h9c9d46e_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-aarch64::rmm-22.08.00-py39hebf1436_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
Downloading: https://conda.anaconda.org/conda-forge/linux-ppc64le/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/linux-ppc64le/repodata.json.bz2
linux-ppc64le::cuda-python-11.7.1-py310h75ac704_0.tar.bz2
-    "python_abi 3.10.* *_cp310"
+    "python_abi 3.10.* *_cp310",
+    "cuda-python >=11.7.1,<12.0a0"
linux-ppc64le::cuda-python-11.7.1-py37h87fa38d_0.tar.bz2
-    "python_abi 3.7.* *_cp37m"
+    "python_abi 3.7.* *_cp37m",
+    "cuda-python >=11.7.1,<12.0a0"
linux-ppc64le::cuda-python-11.7.1-py38h2c9e25b_0.tar.bz2
-    "python_abi 3.8.* *_cp38"
+    "python_abi 3.8.* *_cp38",
+    "cuda-python >=11.7.1,<12.0a0"
linux-ppc64le::cuda-python-11.7.1-py39hfe0bb66_0.tar.bz2
-    "python_abi 3.9.* *_cp39"
+    "python_abi 3.9.* *_cp39",
+    "cuda-python >=11.7.1,<12.0a0"
linux-ppc64le::numba-0.55.2-py310h782a0c3_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-ppc64le::numba-0.55.2-py37h483c640_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-ppc64le::numba-0.55.2-py38h4c6250a_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
linux-ppc64le::numba-0.55.2-py39ha474065_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
linux-ppc64le::rmm-22.06.00-py310hadc03bd_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-ppc64le::rmm-22.06.00-py38habd7850_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-ppc64le::rmm-22.06.00-py39hb0a7ebe_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7,<11.7.1.0a0",
linux-ppc64le::rmm-22.08.00-py310hadc03bd_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-ppc64le::rmm-22.08.00-py38habd7850_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
linux-ppc64le::rmm-22.08.00-py39hb0a7ebe_0.tar.bz2
-    "cuda-python >=11.7,<12.0a0",
+    "cuda-python >=11.7.1,<12.0a0",
Downloading: https://conda.anaconda.org/conda-forge/osx-64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/osx-64/repodata.json.bz2
osx-64::numba-0.55.2-py310hab1bff6_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.55.2-py37h7687b59_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.55.2-py38h1f551d2_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
osx-64::numba-0.55.2-py39hc37845d_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py310h62db5c2_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py310h62db5c2_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py37h7da6166_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py37h7da6166_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py38hab356c4_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
osx-64::numba-0.56.2-py38hab356c4_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py39hf6b9d82_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-64::numba-0.56.2-py39hf6b9d82_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
Downloading: https://conda.anaconda.org/conda-forge/osx-arm64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/osx-arm64/repodata.json.bz2
osx-arm64::numba-0.55.2-py310ha67b2a8_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.55.2-py38h25e2f74_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.55.2-py39hc0d1af8_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py310h3124f1e_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py310h3124f1e_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py38h947a10e_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py38h947a10e_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
osx-arm64::numba-0.56.2-py39h251cc7c_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
osx-arm64::numba-0.56.2-py39h251cc7c_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
Downloading: https://conda.anaconda.org/conda-forge/win-32/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/win-32/repodata.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/win-64/repodata_from_packages.json.bz2
Downloading: https://conda.anaconda.org/conda-forge/win-64/repodata.json.bz2
win-64::cuda-python-11.7.1-py310h872c6dc_0.tar.bz2
-    "vs2015_runtime >=14.16.27033"
+    "vs2015_runtime >=14.16.27033",
+    "cuda-python >=11.7.1,<12.0a0"
win-64::cuda-python-11.7.1-py37hb040724_0.tar.bz2
-    "vs2015_runtime >=14.16.27033"
+    "vs2015_runtime >=14.16.27033",
+    "cuda-python >=11.7.1,<12.0a0"
win-64::cuda-python-11.7.1-py38h4f32bc5_0.tar.bz2
-    "vs2015_runtime >=14.16.27033"
+    "vs2015_runtime >=14.16.27033",
+    "cuda-python >=11.7.1,<12.0a0"
win-64::cuda-python-11.7.1-py39h622b2a6_0.tar.bz2
-    "vs2015_runtime >=14.16.27033"
+    "vs2015_runtime >=14.16.27033",
+    "cuda-python >=11.7.1,<12.0a0"
win-64::numba-0.55.2-py310h77579ad_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.55.2-py37hc29f945_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.55.2-py38h860efb6_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.55.2-py39hb8cd55e_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py310h19bcfe9_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py310h19bcfe9_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py37h9c05cb2_0.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
win-64::numba-0.56.2-py37h9c05cb2_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py38h3873db4_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py38h3873db4_1.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py39h99ae161_0.tar.bz2
-    "cuda-python >=11.6",
+    "cuda-python >=11.6,<11.7.1",
win-64::numba-0.56.2-py39h99ae161_1.tar.bz2
-    "cuda-python >=11.6"
+    "cuda-python >=11.6,<11.7.1"
```

</p>
</details>

recipe/gen_patch_json.py
```diff
@@ -808,6 +808,27 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
...
+            and pkg_resources.parse_version(record["version"]) <= pkg_resources.parse_version("0.56.2")):
+            for i, dep in enumerate(record.get("constrains", [])):
+                if dep.startswith("cuda-python"):
+                    record["constrains"][i] = "cuda-python >=11.6,<11.7.1"
+                    break
+        if (record_name == "cuda-python"
+            and pkg_resources.parse_version(record["version"]) == pkg_resources.parse_version("11.7.1")
+            and record['build_number'] == 0):
+            # add run_exports
+            deps.append("cuda-python >=11.7.1,<12.0a0")
```
leofang, line 830: Q: Does this achieve this desired effect?
```yaml
run_exports:
  - {{ pin_subpackage('cuda-python') }}
```

isuruf: No, you cannot patch `run_exports`. You have to fix it in the recipe for the old version (11.7.0) and the new version (11.7.1)

leofang: Thanks, @isuruf. We're fixing the new version now (https://github.com/conda-forge/cuda-python-feedstock/pull/17), but I don't think we can afford to rebuild all prior versions?

recipe/gen_patch_json.py
```diff
@@ -2112,7 +2128,7 @@ def _relax_libssh2_1_x_pinning(fn, record):
         depends[dep_idx] = "libssh2 >=1.8.0,<2.0.0a0"
 
 
-cb_pin_regex = re.compile(r"^>=(?P<lower>\d(\.\d+)*a?),<(?P<upper>\d(\.\d+)*)a0$")
+cb_pin_regex = re.compile(r"^>=(?P<lower>\d{1,2}(\.\d+)*a?),<(?P<upper>\d{1,2}(\.\d+)*)a0$")
```
kkraus14, line 2131: What is this regex change?

kkraus14 (approved):

recipe/gen_patch_json.py
jakirkham: We could patch packages that depend on `cuda-python` to adjust their version requirement

leofang: @jakirkham that's exactly what I did with this PR.

isuruf: You need to rebuild at least 11.7.0 with the correct run_exports. Otherwise this patch will affect packages built in the future, which we dont want.

recipe/gen_patch_json.py
```diff
@@ -808,6 +808,22 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if i >= 0:
             deps[i] = "cudatoolkit >=11.2,<12.0a0"
 
+        # cuda-python 11.7.1 is ABI-breaking, currently it impacts rmm & numba
+        if record_name == "rmm":
```
jakirkham, line 812: Why are we checking only `rmm`? Potentially there could be other affected packages.

Would suggest following a pattern like [this one]( https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/ebab36266e6619116e8322c3e0cf8908ed4162ea/recipe/gen_patch_json.py#L768-L770 ).

Also we may want to [bound this with a timestamp]( https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/ebab36266e6619116e8322c3e0cf8908ed4162ea/recipe/gen_patch_json.py#L774 ) so we limit this to the packages preceding the `run_export`.

leofang: I determined the impact by searching all conda-forge codebase for `cuda-python`, and concluded that only `rmm` and `numba` are impacted. Wouldn't using precise targets as opposed to a general for loop + if checks reduce the bot's runtime?

recipe/gen_patch_json.py
leofang: @isuruf I do not understand how future packages (`rmm`? `numba`? `cuda-python`?) could be impacted, could you elaborate? Thanks.

recipe/gen_patch_json.py
jakirkham: Not really. The patching occurs as part of the repodata package build in this feedstock. The result is a generated JSON file that we ship. So bots don't need to regenerate this (they just read that JSON).

As part of the build in this feedstock, we loop over all the packages anyways. So this only comes down to how much work it is to check for the package, which is pretty trivial. The timestamp further filters this down. The bigger time sink is if we get this wrong and have to iterate on changing this again.

kkraus14: @jakirkham I don't think there's a way to differentiate between packages that only use the cuda-python Python API vs use the Cython API. If only using the Python API then you can float between versions without issue,

jakirkham: What issue occurs if we pin both of them?

My concern is this filter feels fragile. Spending time doing this by hand is also cumbersome. It may also entail repeated intervention if we missed things.

If this issue occurs again in another version, we may want to borrow from the current code, but as it is formulated we can't do that.

leofang: @jakirkham Unfortunately the suggested pattern does not work for `rmm` because only one version (22.08) was built with CUDA Python 11.7.1. As a result, it must be pinned differently from prior versions. 

For `numba`, I think it only uses CUDA Python's Python APIs (@gmarkall to confirm) so we probably don't even need to patch it.

jakirkham: Sorry, but am not following.

Why are we adding the `run_export` line to `cuda-python` for all packages it builds with ( https://github.com/conda-forge/cuda-python-feedstock/pull/17 ), if we only want this constraint applied to some packages? It seems if we want this applied to only some packages, those packages should be adding `pin_compatible` instead (like what we do with `numpy`).

Please clarify if I'm missing something (again have been out for a while and am only recently catching up on this discussion)

leofang: Hi @jakirkham We added `run_export` to the `cuda-python` recipe because CUDA Python 11.7.0 and 11.7.1+ are ABI incompatible. `pin_compatible` in each individual downstream works, but only in each individual downstream feedstock that depends on `cuda-python`, and they need to be aware of this ABI problem first. I think it is too much to ask.

We do NOT only want this constraint applied to some packages. Again, the patch is done here in this particular way because only two packages are impacted by the ABI incompatibility issue.

gmarkall: > For `numba`, I think it only uses CUDA Python's Python APIs (@gmarkall to confirm) so we probably don't even need to patch it.

I only just noticed this, apologies - indeed Numba only uses the Python API.

leofang: Thanks, Graham. Removed the Numba patch in e66febc.

PR #327, opened 2022-10-06
Remove broken moviepy 1.0.3 build

alanhdu:
This package had an incorrect `.patch` version applied, causing it to fail. I have a fix for the patch file (https://github.com/conda-forge/moviepy-feedstock/pull/17), but it's taking a while to get merged so let's just yank the broken build for now to prevent people from running into this problem.

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -31,6 +31,7 @@ SUBDIRS = (
 
 REMOVALS = {
     "noarch": (
+        "moviepy-1.0.3-pyhd8ed1ab_0.tar.bz2",
         "sendgrid-5.3.0-py_0.tar.bz2",
     ),
     "linux-64": (


ocefpaf:
@alanhdu the right procedure here is to mark that package as broken. Please see https://conda-forge.org/docs/maintainer/updating_pkgs.html#removing-broken-packages

alanhdu:
Ah apologies! I got confused from the admin-request README, which says:

> We prefer to patch the repo data (see [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock)) instead of marking packages as broken. This alternative workflow makes environments more reproducible.

Do you mind clarifying which broken packages should be made in admin-request vs made here? I'd be happy to update the READMe to avoid future confusion.

jakirkham:
Thanks Alan! 🙏

Yeah repodata patches are intended for things where the package is functional, but maybe its dependencies need updating (wrong version constraints, missing dependency, etc.). IOW these are metadata changes about the package.

For packages that are just not functional (as is the case here), we should just remove the package so users can't install it.

Fortunately it is pretty easy to do the latter. If you run into any questions, happy to answer them 🙂

PR #331, opened 2022-10-11
Add missing gcsfs dependency to minimalkv

xhochy:
The 1.4.0 release is missing a dependency on gcsfs. The build was already updated via https://github.com/conda-forge/minimalkv-feedstock/pull/13, this adds it to the existing build 0.

Diff: https://gist.github.com/xhochy/fdc5038f97a9731780862e3bb65603c8


isuruf:
Adding missing deps using repodata patching will lead to loud inconsistency warnings on existing environments. It's better to mark them as broken.

xhochy:
OK, then made this instead: https://github.com/conda-forge/admin-requests/pull/495

PR #341, opened 2022-10-21
Fix dependencies for switch_model 2.0.4 and 2.0.5

mfripp:
Checklist


hmaarrfk:
Is this PR still relevant 2 years lateR?

mfripp:
I don't think so. I may have tried this as one way to update the metadata for this package, then instead switched over to updating it via https://github.com/mfripp/switch_model-feedstock.git.

PR #357, opened 2022-11-17
Update bokeh restrictions on old versions of panel.

erykoff:
Checklist

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -1070,6 +1070,35 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
             deps.append("numba !=0.54.0")
             record["depends"] = deps
 
+        # Patch bokeh version restrictions on older panels.
+        if record_name == "panel":
+            deps = record.get("depends", [])
+            bokeh_dep = None
+            if record["version"] in ["0.1.2", "0.1.3"]:
+                bokeh_dep = "bokeh ==0.12.15"
+            elif record["version"] in ["0.3.1", "0.4.0"]:
+                bokeh_dep = "bokeh >=1.0.0,<1.1.0"
+            elif record["version"] in ["0.5.1", "0.6.0"]:
+                bokeh_dep = "bokeh >=1.1.0,<1.2.0"
+            elif record["version"] in ["0.6.2", "0.6.3", "0.6.4"]:
+                bokeh_dep = "bokeh >=1.3.0,<1.4.0"
+            elif record["version"] in ["0.7.0"]:
+                bokeh_dep = "bokeh >=1.4.0,<1.5.0"
+            elif record["version"] in ["0.9.1", "0.9.2", "0.9.3", "0.9.4", "0.9.5"]:
+                bokeh_dep = "bokeh >=2.0,<2.1"
+            elif record["version"] in ["0.9.6", "0.9.7"]:
+                bokeh_dep = "bokeh >=2.1,<2.2"
+            elif record["version"] in ["0.10.0", "0.10.1", "0.10.2", "0.10.3"]:
+                bokeh_dep = "bokeh >=2.2,<2.3"
+            if bokeh_dep:
+                deps = record.get("depends", [])
+                ind = [deps.index(dep) for dep in deps if dep.startswith("bokeh")]
+                if len(ind) == 1:
+                    deps[ind[0]] = bokeh_dep
+                else:
+                    deps.append(bokeh_dep)
+                record["depends"] = deps
+
         llvm_pkgs = ["clang", "clang-tools", "llvm", "llvm-tools", "llvmdev"]
         if record_name in llvm_pkgs:
             new_constrains = record.get('constrains', [])


beckermr:
cc @conda-forge/panel

erykoff:
FYI, the motivation for this PR is that the solver for `rubin-env=5.0.0` is giving `bokeh=3` and `panel=0.10.3` because they are (according to the pins) compatible, but they actually aren't.  We can certainly force `bokeh=2` (or `panel=0.14`) which will get compatible versions, but wanted to fix this upstream for everybody.

beckermr:
Where did the rest of the constraints come from?

erykoff:
Also, the version pins were chosen to match the lower pins.  For example if the pin was `bokeh >=2.0` I thought the conservative choice would be to pin to `bokeh >=2.0,<2.1`.  I ran `gen_patch_json.py` locally to confirm that my updates were compatible with the original pins.

maximlt:
Interesting @erykoff thanks for starting to address that! This is biting Panel users more and more (and HoloViews/hvPlot users, as these tools depend on Panel).

I think we need to discuss more about it, one thing I'm concerned with is the possibility of these tighter pins to break old environments. We may need to put in more loose pins instead.

erykoff:
If you have any ideas what the max pins should be, please chime in.  I certainly don't have any knowledge of this, or ability to test.

philippjfr:
@erykoff Really, really appreciate your efforts here. I double checked all the pins here and they look correct to me. Panel should have pinned Bokeh tightly from the very beginning but I was sufficiently aware of the lack of stability of the Bokeh.js APIs. Overall I'm in favor of merging this, the early versions of Panel may be more forgiving in their Bokeh.js compatibility because the number of JS models was so small which means the API surface is probably pretty small and forgiving. However early versions of Panel will also have had a pretty small userbase and all versions >0.7 or so really did require quite strict pins.

I'm +1 on merging this.

ocefpaf:
> I'm +1 on merging this.

If the upstream author is fine with this let's merge! Thanks @philippjfr for the assessment and @erykoff for the PR.

timkpaine:
This is a disaster. You cannot just rewrite metadata for old versions without doing a newer build. This breaks every solved environment.

philippjfr:
@timkpaine thanks for chiming in. I was under the impression that metadata patches like this could be applied in cases where old versions were insufficiently pinned (which is the case here) and resulted in broken environments. I very much appreciate the concern about breaking old environments, however releasing new builds for all old versions with insufficient pins would be quite onerous in this case.

erykoff:
There were no patches for anything that already had an upper bound. Hopefully it didn’t break everything…

philippjfr:
Thanks @erykoff, in that case we really only had two options. Kick off 21 new builds or what we did here.

ocefpaf:
> Thanks @erykoff, in that case we really only had two options. Kick off 21 new builds or what we did here.

21 new builds that would not be equal b/c conda-forge's build system does not ensure build reproducibility. Do you had only 1 choice, what you did here ;-p

(Unless the metadata correction is not required. But reading the issue and this PR it seems that it was needed.)

timkpaine:
Or do let's say 1-2 new builds for latest releases and leave 2 year old releases alone

erykoff:
Leaving the old releases alone makes them get installed with new bokeh which is the problem this is solving.

philippjfr:
>Or do let's say 1-2 new builds for latest releases and leave 2 year old releases alone.

That wouldn't solve the issue though because the solver would have simply gone back to the most recent release without an upper pin.

ocefpaf:
> > Or do let's say 1-2 new builds for latest releases and leave 2 year old releases alone.
> 
> That wouldn't solve the issue though because the solver would have simply gone back to the most recent release without an upper pin.

Exactly. That is why we ended up with the model of repodatapatches, to ensure correctness and recommend users that require absolute reproducibility to lock their envs with a lockfile. Which won't be affected by the patches.

timkpaine:
> That wouldn't solve the issue though because the solver would have simply gone back to the most recent release without an upper pin.

For people who don't specify a min version, which is the behavior you'd want if you didn't want to break people with old stuff. At the end of the day it's fine I guess, the lesson learned is follow this repo very closely as big changes get pushed to prod with little oversight, and don't trust that panel is ready for prime time. ¯\_(ツ)_/¯

philippjfr:
I can't argue If that's your takeaway. However, these patches are fixing mistakes with version pins that happened over two years ago, which is exactly what this patch system is meant to address.

ocefpaf:
> As it may be tricky to generate a lockfile at this point (given the repodata has already changed on Anaconda.org), it is worth noting that one can download the unpatched repodata and [use this repodata file locally](https://docs.conda.io/projects/conda/en/stable/commands/create.html#Channel%20Customization).

We should probably add that to our docs too to help users figure out their alternatives.

timkpaine:
> As it may be tricky to generate a lockfile at this point (given the repodata has already changed on Anaconda.org), it is worth noting that one can download the unpatched repodata and [use this repodata file locally](https://docs.conda.io/projects/conda/en/stable/commands/create.html#Channel%20Customization).

I'll look into this. I pinned down to a hash/build version which unfortunately doesn't work anymore and breaks the env solving from then on.

jakirkham:
FWIW `conda list` supports an `--explicit` flag, which creates a specification file that pins packages down to exact URLs. This might be another approach

timkpaine:
> FWIW `conda list` supports an `--explicit` flag, which creates a specification file that pins packages down to exact URLs. This might be another approach

This is great, is there a way to do this in a recipe? I use `pin_depends: strict` which isn't strict enough

timkpaine:
Or maybe I can go and implement `pin_depends: url` for conda build

erykoff:
If you know that the pins here are too strict for a particular old panel (due to your tests) I think you can patch again in this repo. Certainly bokeh 3 won’t work!

timkpaine:
> If you know that the pins here are too strict for a particular old panel (due to your tests) I think you can patch again in this repo. Certainly bokeh 3 won’t work!

Anything other than undoing the changes will break me, I have assets that `pin_depends: strict` which resolved to a certain panel and bokeh. My only option is to regenerate. I was used to yanking as a breakage vector but this PR introduced a new, very widely scoped breakage vector (since yank usually is only new stuff, old stuff only for security issues usually).

philippjfr:
>I have assets that pin_depends: strict which resolved to a certain panel and bokeh

If it's not too much trouble, what versions did they resolve to? I'm guessing it's Bokeh 3.0.x and Panel 0.10.3, which is not a supported combination. I appreciate that's probably not a huge concern for you since you're already happy to drop Panel. If there is some mechanism that will not result in new users getting a broken set of packages but also maintains backwards compatibility for older solves I'd be very happy to put effort into making that happen.

jbednar:
> big changes get pushed to prod pretty easily (I'm watching all changes now), and don't trust that panel is stable.

For clarity, Panel hasn't changed at all; the only thing that has happened is that Bokeh 3 includes breaking changes (hence moving from 2.x to 3 in semantic versioning), and is not compatible with any existing version of panel.  Updating the repodata accurately reflects this actual situation, and should not break any environment that actually works. If some environment had any current version of panel alongside bokeh 3, it was not an environment where panel could actually be used.

PR #358, opened 2022-11-18
add repodata patch for iris with `netCDF4<1.6.0` pin

valeriupredoi:
Checklist

Here is the output for `show_diff.py`
[show_diff_PR358.txt](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/files/10043065/show_diff_PR358.txt)

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -815,7 +815,7 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if record_name == "iris":
             _rename_dependency(fn, record, "nc_time_axis", "nc-time-axis")
             if record["version"] in iris_updates:
-                record["depends"] = iris_updates[record["version"]]
+                record["depends"].extend(iris_updates[record["version"]])
 
         if record_name == "nordugrid-arc" and record.get("timestamp", 0) < 1666690884000:
             record["depends"].append("glibmm-2.4 >=2.58.1")
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -795,8 +795,39 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if record_name == "airflow-with-async":
             _rename_dependency(fn, record, "evenlet", "eventlet")
 
+        # Current implementation in iris allows for frequent SegFaults
+        # when netCDF4>1.6.0; iris devs are working on a solution and
+        # discussing it in https://github.com/SciTools/iris/issues/5016
+        # but a wide repodata patch is needed before a permanent solution
+        iris_deps = [
+            "python >=3.8",
+            "pip",
+            "setuptools >=64",
+            "setuptools-scm >=7",
+            "python >=3.8",
+            "cartopy >=0.21",
+            "cf-units >=3.1",
+            "cftime >=1.5",
+            "dask-core >=2",
+            "matplotlib-base",
+            "netcdf4 <1.6.1",
+            "numpy >=1.19",
+            "pyproj",
+            "scipy",
+            "shapely !=1.8.3",
+            "python-xxhash"
+        ]
+        iris_updates = {
+            "3.1.0": iris_deps,
+            "3.2.0": iris_deps,
+            "3.2.1": iris_deps,
+            "3.3.0": iris_deps,
+            "3.3.1": iris_deps,
+            "3.4.0": iris_deps,
+        }
         if record_name == "iris":
             _rename_dependency(fn, record, "nc_time_axis", "nc-time-axis")
+            record["depends"] = iris_updates[record["version"]]
 
         if record_name == "nordugrid-arc" and record.get("timestamp", 0) < 1666690884000:
             record["depends"].append("glibmm-2.4 >=2.58.1")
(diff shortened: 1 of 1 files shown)


valeriupredoi:
@bjlittle @zklaus @bouweandela here is a repodata patch that we believe we should get done before things gets very frustrating given the situation with netCDF>1.6.0 and iris. Please be gentle, since this is my first repodata patch PR :grin: Specifically: I am not 100% sure of the iris versions we need to patch, please have a look :beer:

valeriupredoi:
OK I have now restricted versions to newer than (inclusive of) iris=3.1.0 (that's pretty old anyway)

trexfeathers (commented):
Good idea, thanks very much @valeriupredoi! The Iris core devs are happy with this in principle. Hopefully @conda-forge/core can say whether it is written correctly.

valeriupredoi:
@trexfeathers brilliant, many thanks! (I only now got to understand your GH handle, when I had to write it w/o autocomplete :laughing: :t-rex: )

recipe/gen_patch_json.py
```diff
@@ -795,8 +795,27 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
...
+            "3.2.1": iris_deps,
+            "3.2.1.post0": iris_deps,
+            "3.3.0": iris_deps,
+            "3.3.1": iris_deps,
+            "3.4.0": iris_deps,
+        }
         if record_name == "iris":
             _rename_dependency(fn, record, "nc_time_axis", "nc-time-axis")
+            if record["version"] in iris_updates:
+                record["depends"] = iris_updates[record["version"]]
```
beckermr, line 818: This overwrites all of the dependencies. I think you want to append or edit them, right?

valeriupredoi: does it? Yeah all I want is to pin netCDF4 <1.6.1 for all those listed versions. But looking at the show_diff output and the test it looks to me that the deps and their pins are preserved as they are for each version, and only netCDF4 <1.6.1 is appended at the end of the list?

beckermr: Not it removes everything. Notice the `-` on the diff

```diff
noarch::iris-3.3.1-pyhd8ed1ab_0.tar.bz2
-    "cartopy >=0.21",
-    "cf-units >=3.1",
-    "cftime >=1.5",
-    "dask-core >=2",
-    "matplotlib-base",
-    "netcdf4 !=1.6.1",
-    "numpy >=1.19",
-    "pyproj",
-    "python >=3.8",
-    "python-xxhash",
-    "scipy",
-    "shapely !=1.8.3"
+    "netcdf4 <1.6.1"
```

valeriupredoi: crap! Nevermind me, I am a total doofus, I only now noticed the output is an actual diff, fixing it now, thanks a lot @beckermr

valeriupredoi: totally! I should pay attention to signs at the start of the line :laughing: It now should do what's written on the tin, let me upload the diff output to the PR description again

recipe/gen_patch_json.py
```diff
@@ -815,7 +815,7 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if record_name == "iris":
             _rename_dependency(fn, record, "nc_time_axis", "nc-time-axis")
             if record["version"] in iris_updates:
-                record["depends"] = iris_updates[record["version"]]
+                record["depends"].extend(iris_updates[record["version"]])
```
beckermr, line 818: You should find the current netcdf pin and replace it like other examples in the file.

valeriupredoi: I was thinking I should do that - but no version pins it until 3.4 (which is coming now, it's in rc0 stage at the moment), and that pins it to !=1.6.1 so am afraid to mess around with the various ways of not pinning/pinning it - wouldn't the append at the end of the deps list do the trick in a safe way?

beckermr: That's fine.

valeriupredoi: yuss! Cheers muchly @beckermr very much appreciated :beer:

recipe/gen_patch_json.py
```diff
@@ -795,8 +795,27 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if record_name == "airflow-with-async":
             _rename_dependency(fn, record, "evenlet", "eventlet")
 
+        # Current implementation in iris allows for frequent SegFaults
+        # when netCDF4>1.6.0; iris devs are working on a solution and
+        # discussing it in https://github.com/SciTools/iris/issues/5016
+        # but a wide repodata patch is needed before a permanent solution
```
valeriupredoi, line 801: @trexfeathers maybe this message should be changed to "We darn well fixed it - but it's still safe to keep the patch for iris <3.4.1" - up to you if you wanna change, in a simple PR

trexfeathers: Yeah this definitely looks like I should update it. Thanks!

trexfeathers: #402

PR #362, opened 2022-11-30
Exclude libxml2 2.9.11 and 2.9.12 in lxml

beenje:
This is an attempt to fix https://github.com/conda-forge/lxml-feedstock/issues/75

libxml2 2.9.11 and 2.9.12 versions have a bug: https://bugs.launchpad.net/lxml/+bug/1928795
This was fixed in libxml feedstock 4.8.0 build 3: https://github.com/conda-forge/lxml-feedstock/pull/71

But previous builds are currently still selected by conda/mamba.
I applied the patch to all builds prior to 4.8.0 build 3. That's many packages. I couldn't paste the diff here. So I created a gist: https://gist.github.com/beenje/dbfcd01af4e69dac7ee480685b3c073b

Maybe applying the patch only for 4.8.0 (build 0, 1 and 2) would be enough?

ping @ocefpaf

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -1705,6 +1705,13 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
         if record_name == "guiqwt" and record["version"] in ("3.0.5", "3.0.7") and record["build_number"] in (0, 1):
             _replace_pin("qtpy >=1.3", "qtpy >=1.3,<2.0", record["depends"], record)
 
+        # exclude libxml2 2.9.11 and 2.9.12 due to https://bugs.launchpad.net/lxml/+bug/1928795
+        # https://github.com/conda-forge/lxml-feedstock/issues/75
+        if record_name == "lxml" and record.get("timestamp", 0) <= 1649691000000: # 4.8.0 build 2 and prior
+            for i, dep in enumerate(record["depends"]):
+                if dep.startswith("libxml2 >=2.9."):
+                    record["depends"][i] = f"{dep},!=2.9.11,!=2.9.12"
+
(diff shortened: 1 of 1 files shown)


Guspyro (approved):

ocefpaf:
@isuruf I know you are terribly busy but can you take a quick look at this PR?

basnijholt:
This PR breaks all of the environment files we have for our internal software. 

We also build conda packages where we strictly pin each version. Now we cannot install any of the builds of the last year anymore.

Is this really desired behavior?

ocefpaf:
> Is this really desired behavior?

If my understanding of the issue is correct that `libxml2` version is problematic with `lxml` and this is probably the best solution we can afford without creating broken envs. However, maybe we don't have all the info on how those versions play together.

hmaarrfk:
@basnijholt do you pin versions, or also hashes?

Generally speaking we have been patching out metadata to help users that create new environments avoid software that we find to be incompatible after it has been released.

If you pin hashes too, maybe there are ways to "ignore" the metadata, though I"m not super familar with the syntax and flags you would need to pass to conda/mamba.

isuruf:
> this is probably the best solution we can afford without creating broken envs.

https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/370 should fix the original issue.
If you want to be extra careful, you can build a new `libxml2=2.9.14` with `icu=69`.

isuruf:
Now that #370, we can revert this unless someone wants to do a `libxml2=2.9.14` build with `icu=69` before doing that.

basnijholt:
Thank you so much for fixing the problem properly.

Reverting would be great! I’ve had several people contact me already and at least a dozen CI pipelines are broken 😅

I will open a PR.

PR #471, opened 2023-06-19
pin pluggy<1.1.0 in conda

jaimergp:
Checklist

pluggy 1.1.0 released today breaks `conda` plugin system and prevents its usage.  See https://github.com/conda/conda/issues/12808


recipe/gen_patch_json.py
```diff
@@ -2348,7 +2352,7 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
                         record,
                     )
                 else:
-                    # old versions depended on urllib3 via requests;
+                    # old versions depended on urllib3 via requests;
```
jaimergp, line 2355:
```suggestion
                    # old versions depended on urllib3 via requests;
```

PR #473, opened 2023-06-21
patch hdf5 1.14.0 builds to accept 1.14.1

minrk:
ABI did not change, we are still mid migration

As suggested by @hmaarrfk in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4594

Checklist

Note: I don't think this will actually work, and we should go with https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4594 instead, because a migrator still needs to be applied to all repos already migrated to start building with 1.14.1 instead of 1.14.0, which won't get patched due to the timestamp filter.

We can still apply this patch to accept the old builds, but it does not _replace_ the 1.14.1 migration, which is still required.


minrk:
[diff in a gist](https://gist.github.com/minrk/29082115ebca258b71981b9bdf99ef57) because GitHub rejected it in the PR/comment body as too long.

hmaarrfk:
sorry i missed this. 

with the new organization, this would have to be rewritten.

PR #478, opened 2023-07-11
Pin python>=3.8 for zipp 3.16.0

dtreichler:
Checklist

```
noarch::lbnightlytools-3.0.56-pyhd8ed1ab_0.conda
-    "python =2.7|>=3.6",
+    "python ==2.7.*|>=3.6",
noarch::zipp-3.16.0-pyhd8ed1ab_0.conda
-    "python >=3.7"
+    "python >=3.8"
```

(n.b., the `lbnightlytools` changes appear to be related to #444)

See conda-forge/zipp-feedstock#43


h-vetinari (approved):
I noted this in the zipp-PR already - Python 3.7 has been dropped for almost a year, and must be considered completely unsupported because we have zero testing for it anymore, and what's not tested is generally broken.

That said, I'm okay with the fix in this case, and to unbreak people affected by this.

PR #481, opened 2023-07-12
codespell: config, workflow + typos fixed

yarikoptic:
to ensure absence of typos


beckermr (changes requested):
We do not allow adhoc workflows in our feedstocks and smithy rerenders will delete them. Thus I don't think we add this in.

yarikoptic:
ah, ok. Do you want me to drop a commit with the workflow and leave the rest in? or just leave the typos fixed in?

beckermr:
Happy to merge the typo fixes! Thank you!

yarikoptic:
ok, kept only last commit

ocefpaf:
We can add a .precommit.yaml file for codespell and run pre-commit-ci here. That would work. But I don't have a strong opinion on this. Documentation should have codespell, this repo, not so much. We can do something like this PR from time to time, no need for strong spell checking.

PR #498, opened 2023-08-15
WIP yaml based-patch system because this repo broke me

beckermr:
Checklist

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -15,6 +15,8 @@ import packaging.version
 import pkg_resources
 
 from get_license_family import get_license_family
+from patch_yamls_utils import patch_yamls_edit_index
+
 
 CHANNEL_NAME = "conda-forge"
 CHANNEL_ALIAS = "https://conda.anaconda.org"
@@ -434,6 +436,7 @@ def _gen_new_index(repodata, subdir):
     indexes = {}
     for index_key in ['packages', 'packages.conda']:
         indexes[index_key] = _gen_new_index_per_key(repodata, subdir, index_key)
+        patch_yamls_edit_index(indexes[index_key], subdir)
 
     return indexes
 
@@ -499,14 +502,6 @@ def _gen_new_index_per_key(repodata, subdir, index_key):
             family = get_license_family(record["license"])
             if family:
                 record['license_family'] = family
-
-        # test dot-conda patches
-        if (
-            record_name == "cf-autotick-bot-test-package"
-            and record["version"] == "0.14"
-            and fn.rsplit(".", 1)[0].endswith("_2")
-        ):
-            record["depends"].append("tqdm")
 
         # remove dependency from constrains for twisted
         if record_name == "twisted":
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -30,6 +30,7 @@ test:
     - gen_patch_json.py
     - get_license_family.py
     - show_diff.py
+    patch_yamls_utils.py
   requires:
     - python 3.*
     - setuptools <66.0.0a
@@ -38,10 +39,14 @@ test:
     - license-expression
     - pytest
     - conda-build
+    - black
+    - ruff
   commands:
     {% for subdir in ("noarch", "linux-64", "linux-armv7l", "linux-ppc64le", "osx-64", "osx-arm64", "win-32", "win-64") %}
     - test -e $PREFIX/{{ subdir }}/patch_instructions.json
     {% endfor %}
+    - black --check patch_yamls_utils.py
+    - ruff patch_yamls_utils.py
     - pytest -vv test_gen_patch_json.py
     - python show_diff.py
 
--- a/recipe/patch_yamls/cf-autotick-bot-test-package.yaml
+++ b/recipe/patch_yamls/cf-autotick-bot-test-package.yaml
@@ -1,0 +1,7 @@
+# test dot-conda patches
+where:
+  name: cf-autotick-bot-test-package
+  version: "0.14"
+  build_number: 2
+do:
+  add_depends: tqdm
--- a/recipe/patch_yamls_utils.py
+++ b/recipe/patch_yamls_utils.py
@@ -1,0 +1,65 @@
+import yaml
+import glob
+import os
+
+
+ALL_YAMLS = []
+for fname in glob.glob(os.path.dirname(__file__) + "/patch_yamls/*.yaml"):
+    with open(fname, "r") as fp:
+        fname_yamls = [
+            patch_yaml
+            for patch_yaml in yaml.safe_load_all(fp)
+            if patch_yaml is not None
+        ]
+        ALL_YAMLS += fname_yamls
+        print(
+            "Read %d patch yamls for file %s"
+            % (
+                len(fname_yamls),
+                os.path.basename(fname),
+            ),
+            flush=True,
+        )
+
+print("Read %d total patch yamls" % len(ALL_YAMLS), flush=True)
+
+
+def _apply_patch_yaml(patch_yaml, record, subdir, fn):
+    keep = True
+    for k, v in patch_yaml["where"].items():
+        if k == "max_timestamp":
+            keep = keep and (record.get("timestamp", 0) < v)
+            if not keep:
+                break
+        elif k == "version_in":
+            keep = keep and (record["version"] in v)
+            if not keep:
+                break
+        elif k == "subdir":
+            keep = keep and (subdir == v)
+            if not keep:
+                break
+        elif k in record:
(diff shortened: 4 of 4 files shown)


recipe/meta.yaml
```diff
@@ -30,6 +30,7 @@ test:
     - gen_patch_json.py
     - get_license_family.py
     - show_diff.py
+    patch_yamls_utils.py
```
beckermr, line 33:
```suggestion
    - patch_yamls_utils.py
```

beckermr:
@conda-forge-admin rerender

beckermr:
OK @conda-forge/core. I finally got so frustrated with this repo that I broke down and wrote a small YAML-based system for common patch patterns. I don't have a ton of more energy for this and would like to see this merged since we absolutely need it to manage the complexity here. 

So please add your comments, keeping in mind that I will very likely not make any large-scale changes to the code due to time. If you have strong feelings that this cannot be merged without a ton of extra work, then we can let this sit. 

I know the code needs tests. I'll write some basic tests once we have some semblance of agreement on the YAML format and the API. 

See the README.md in the recipe/patch_yaml directory for how it all works.

recipe/patch_yaml/README.md
```diff
@@ -0,0 +1,113 @@
...
+  # possible conditions
+  # list of subdirs or a single subdir (e.g., "linux-64")
+  subdir_in: linux-64
+
+  # list of artifact names or a single name (e.g., "ngmix-2.3.0-py38h50d1736_1.conda")
+  artifact_in: ngmix-2.3.0-py38h50d1736_1.conda
+
+  # any key in the repodata entry (e.g., "version" or "build_number") with an operation
+  <repodata key>_<ge, gt, le, lt>: <value>
+
```
beckermr, line 30:
```suggestion
```

recipe/patch_yaml/README.md
```diff
@@ -0,0 +1,113 @@
...
+  version_gt: 1.0.0
+
+  # any key in the repodata entry (e.g., "version" or "build_number") and a list of values or single value
+  <repodata key>_in: <list or single item>
+
+  # this means the build number is in the set {0, 1, 2}
+  build_number_in: [0, 1, 2]
+
+  # has specific dependencies as either a list or a single string
+  has_depends: numpy *  # matches any numpy entry with or without a version
```
beckermr, line 41:
```suggestion
  has_depends: numpy *  # matches any numpy entry with or without a version
  has_depends: numpy  # matches "numpy" exactly (i.e., no pins)
```

recipe/meta.yaml
```diff
@@ -30,25 +30,36 @@ test:
...
+    - ruff patch_yaml_utils.py
     - pytest -vv test_gen_patch_json.py
     - python show_diff.py
 
 about:
   summary: generate tweaks to index metadata, hosted separately from anaconda.org index
   home: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock
   license: CC-PDDC
+  description: |
+    See the [README](recipe/patch_yaml/README.md) for how generate patches from static YAML files.
```
beckermr, line 60:
```suggestion
    See the [recipe/patch_yaml/README.md](recipe/patch_yaml/README.md) for how generate patches from static YAML files.
```

beckermr:
@conda-forge-admin rerender

recipe/gen_patch_json.py
```diff
@@ -12,9 +12,20 @@
...
+    patch_yaml_edit_index,
+    _extract_track_feature,
+    _replace_pin,
+    _rename_dependency,
+    _relax_exact,
+    _pin_stricter,
+    _pin_looser,
+    CB_PIN_REGEX,
+    pad_list,
+)
```
jaimergp, line 28: Nice. Most of these were originally in this file but then moved to the utils module, right? I am not sure if that module should be named `patch_yaml_utils` though, since its functions are used in other stuff too.

recipe/patch_yaml/README.md
```diff
@@ -0,0 +1,113 @@
...
+  - relax_exact_depends:
+      # the package name whose constraint should be relaxed
+      old: matplotlib
+      # optional string of 'x', 'x.x' etc. format specify an upper bound
+      # if not given, no upper bound is applied
+      # max_pin: 'x.x'
+
+  # make a dependency version constraint stricter
+  - tighten_depends:
+      # package to pin stricter
```
jaimergp, line 83: To be clear, this is the package _name_ right? Not the full spec. I don't see the `old` vs `new` pairing here, so just double checking. I could have understood this as "gimme the full thing as you find it in the record", so maybe `name` or `pkg_name` would be clearer.

recipe/patch_yaml/README.md
```diff
@@ -0,0 +1,113 @@
...
+Patches are specified by two main blocks.
+
+- The `if` block specifies a set of conditions under which the changes in the `then` block are applied.
+- The different conditions in the `if` block are combined with a logical `AND`.
+- The `if` conditions can use shell glob syntax as implemented in the python `fnmatch` module in the
+  standard library.
+- Multiple patches can be in the same file using separate YAML documents (i.e., separate the data by `---`
+  on a new line).
+
+```yaml
```
jaimergp, line 19: It would be nice to have this as `pydantic` model / JSON schema for easy validation before running the patch. I can contribute one if think it's a good idea!

recipe/patch_yaml_utils.py
```diff
@@ -0,0 +1,310 @@
...
+    return any(fnmatch.fnmatch(item, _v) for _v in v)
+
+
+def _test_patch_yaml(patch_yaml, record, subdir, fn):
+    keep = True
+    for k, v in patch_yaml["if"].items():
+        if k == "subdir_in":
+            keep = keep and _fnmatch_str_or_list(subdir, v)
+            if not keep:
+                break
```
jaimergp, line 43: I think we can move this block to the end of the elif chain (after the else) instead of having it in every condition, given that they all are exclusive.

recipe/patch_yaml_utils.py
```diff
@@ -0,0 +1,310 @@
...
+        ):
+            if max_pin is not None:
+                upper_bound = get_upper_bound(dep_parts[1], max_pin) + "a0"
+                depends[dep_idx] = "{} >={},<{}".format(*dep_parts[:2], upper_bound)
+            else:
+                depends[dep_idx] = "{} >={}".format(*dep_parts[:2])
+            record["depends"] = depends
+
+
+CB_PIN_REGEX = re.compile(r"^>=(?P<lower>\d(\.\d+)*a?),<(?P<upper>\d(\.\d+)*)a0$")
```
jaimergp, line 174: I know this is existing code that moved from the other module, but I wonder how tricky it'd be to use the `MatchSpec` here directly to avoid this kind of regex.

recipe/patch_yaml/README.md
```diff
@@ -0,0 +1,113 @@
...
+  # remove entries from track_features
+  - remove_track_feature: <list of str or str>
+
+  # replace entries via an exact match in either the depends or constrains sections
+  - replace_<depends or constrains>:
+      # str of thing to be replaced
+      old: matplotlib ==1.3.0
+      # thing to replace `old` with
+      new: matplotlib-base ==1.4.0
+      # ^^^^^^ NOTICE the extra indent here!
```
xhochy, line 63: I'm confused by the comment. I would have expected `new` to be then indented one more than `old`.  Maybe skip the comment for clarity?

recipe/patch_yaml/README.md
beckermr: Yes please feel free to contribute directly to this PR!

recipe/patch_yaml_utils.py
beckermr: I am definitely going to punt on this one, but feel free to try!

beckermr: Given how minimal our testing is here and how essential this data is, I am already nervous about big changes.

recipe/gen_patch_json.py
beckermr: My intent is that all patches end up as yaml and so the utility functions are first class and old used in a hacky way in the original code. Thus I think the name (and also the fact that these utils have `_` at the start still) is sensible to me.

beckermr:
OK @jaimergp @xhochy 

I fixed a bug where some keys like timestamp are not in all record. I also addressed the code review where I could and made some of the code less verbose. 

I am going to write some tests since it appears the API is OK for now.

beckermr:
OK @jaimergp @xhochy This one is ready to go from my perspective. @jaimergp You can do the pydantic work here or in a separate PR, your call!

jaimergp (approved):

PR #503, opened 2023-08-17
ENH add pydantic based validation for the new YAML patch instructions

jaimergp:
Follow up to #498 

cc @beckermr

### DIFF
--- a/recipe/patch_yaml_utils.py
+++ b/recipe/patch_yaml_utils.py
@@ -7,10 +7,10 @@ import fnmatch as _fnmatch
 import re
 from functools import lru_cache
 
-ALLOWED_TEMPLATE_KEYS = ["name", "version", "build_number", "subdir"]
-
 from patch_yaml_model import PatchYaml, _IfClause, _ThenClauseItem
 
+
+ALLOWED_TEMPLATE_KEYS = ["name", "version", "build_number", "subdir"]
 
 OPERATORS = ["==", ">=", "<=", ">", "<", "!="]
 
--- a/recipe/patch_yaml_utils.py
+++ b/recipe/patch_yaml_utils.py
@@ -5,6 +5,9 @@ from packaging.version import parse as parse_version
 import fnmatch
 import re
 
+from patch_yaml_model import PatchYaml, _IfClause, _ThenClauseItem
+
+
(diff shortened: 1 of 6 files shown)


recipe/patch_yaml_utils.py
```diff
@@ -260,7 +265,7 @@ def _apply_patch_yaml(patch_yaml, record, subdir, fn):
                 elif not depends and subk in record:
                     del record[subk]
 
-            elif k == "remove_track_feature":
+            elif k == "remove_track_features":
```
jaimergp, line 268: I renamed this field for consistency with the other instructions, where the convention seemed to be `{action}_{repodata-field}`.

beckermr: Yes thank you! This was bothering me yesterday!

beckermr (approved):

beckermr:
This code seems super slow now @jaimergp. Do we have to validate as we patch, or can we only validate in the tests?

beckermr:
Yes, we validate all yamls in the tests that run as part of the package build. This will catch validation issues as PRs are sent.

beckermr:
Shall we merge @jaimergp?

jaimergp:
I think that's ok. I'm a bit puzzled on the performance though. The whole test suite [took 0.69s to run](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=771794&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3&l=180) (and that includes validation, right?), but the patching [still takes ~7 minutes](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=771794&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3&l=197). Not sure that was a bottleneck, but if it works it works!

beckermr:
We were validating each part of each yaml for every yaml and every record in the repodata json. So we were literally running the test suite hundreds of thousands of times.

jaimergp:
> So we were literally running the test suite hundreds of thousands of times.

Ah ok, I see! Thanks for clarifying.

PR #506, opened 2023-08-17
Fix protobuf run dependency on old libignition-msgs5 Windows packages (yaml version)

traversaro:
Fix https://github.com/conda-forge/libignition-msgs1-feedstock/issues/30
Fix https://github.com/gazebosim/gazebo-classic/issues/3339 

Checklist

* [x] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/patch_yaml/README.md)).

### DIFF
--- a/recipe/patch_yaml/libignition-msgs1.yaml
+++ b/recipe/patch_yaml/libignition-msgs1.yaml
@@ -1,0 +1,47 @@
+# Before 2021, libignition-msgs5 had a bug for which no dependency
+# was added on protobuf, and this can create problem when a new protobuf
+# verson is relased, see 
+# https://github.com/conda-forge/libignition-msgs1-feedstock/issues/30
+if:
+  subdir_in: win-64
+  artifact_in: libignition-msgs5-5.3.0-h13ae965_2.tar.bz2
+  timestamp_lt: 1611218248868
+then:
+  - add_depends: "libprotobuf >=3.13.0.1,<3.14.0a0"
+---
+if:
+  subdir_in: win-64
+  artifact_in: libignition-msgs5-5.3.0-h21ff451_1.tar.bz2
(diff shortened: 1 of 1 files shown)


beckermr:
Nice! FYI `artifact_in` can be a list of strings so you can combine some of the patches if you'd like!

traversaro:
> Nice! FYI `artifact_in` can be a list of strings so you can combine some of the patches if you'd like!

Cool! I just did it.

recipe/patch_yaml/libignition-msgs1.yaml
```diff
@@ -0,0 +1,46 @@
...
+if:
+  subdir_in: win-64
+  artifact_in: libignition-msgs5-5.1.0-h21ff451_1.tar.bz2
+  timestamp_lt: 1611218248868
+then:
+  - add_depends: "libprotobuf >=3.11.4,<3.12.0a0"
+---
+if:
+  subdir_in: win-64
+  artifact_in: libignition-msgs5-5.1.0-h21ff451_0.tar.bz2
```
beckermr, line 44: Did you want a timestamp here?

traversaro: Yes, thanks for noticing!

beckermr:
Comments on the api or user experience with the yaml here are appreciated!

traversaro:
> Comments on the api or user experience with the yaml here are appreciated!

It is definitely a big improvements over editing the big python script. I remember that the first time I had to modify the big python script it took me some time to understand were I was supposed to change, while with yaml and one file for feedstock everything is much more clear.

beckermr:
Great!

Paste in the diff when you get a chance and then we can merge if it looks OK.

PR #513, opened 2023-08-22
ENH allow template strings

beckermr:
Checklist


recipe/patch_yaml/README.md
```diff
@@ -61,6 +63,8 @@ then:
   # list of instructions to change things
   # add to the depends or constrains section of the repodata
   - add_<depends or constrains>: <list of str or single str>
+  # you can access the data of the record being patched like this
```
beckermr, line 66:
```suggestion
  # you can use data from the record being patched like this
  # only name, version, build_number and subdir are supported
```

PR #523, opened 2023-08-27
PERF make patches in parallel

beckermr:
Checklist

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -3089,8 +3089,6 @@ def _do_subdir(subdir):
 
 
 def main():
-
-    repodatas = {}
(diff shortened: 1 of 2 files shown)


recipe/gen_patch_json.py
```diff
@@ -3060,41 +3061,50 @@ def _extract_and_remove_vc_feature(record):
...
+        prefix_subdir, "patch_instructions.json")
+    with open(patch_instructions_path, 'w') as fh:
+        json.dump(
+            instructions, fh, indent=2,
+            sort_keys=True, separators=(',', ': '))
+
+
 def main():
+
     # Step 1. Collect initial repodata for all subdirs.
```
beckermr, line 3093:
```suggestion
```

PR #531, opened 2023-08-30
Add a upper bound for rhash since 1.4.4 has SOVERSION change

isuruf:
Checklist


beckermr:
Lgtm! How do you like the new system?

isuruf:
`has_depends` seems a bit limited, but other than that it works great. (Don't have the time to look into it any way)

beckermr:
Ah good feedback. Thank you! 

We support glob syntax in `has_depends` (and basically all of the other selection fields) and one can use `not_*` to negate things. We also support has_constrains. This has covered most patches so far, but not all of them for sure!

PR #535, opened 2023-09-01
ENH more yaml

beckermr:
Checklist


recipe/patch_yaml/singularity-hpc.yaml
```diff
@@ -0,0 +1,17 @@
...
+#     record["depends"].append("jinja2")
+#     record["depends"].append("jsonschema")
+#     record["depends"].append("requests")
+#     record["depends"].append("ruamel.yaml")
+#     record["depends"].append("spython >=0.2.0")
+if:
+  name: singularity-hpc
+  timestamp_lt: 1652410323526
+then:
+  - add_depends: jinja2
```
jaimergp, line 13: I wonder if `{add,remove}_{depends,constrains}` should accept lists too, but I guess it's mostly cosmetic.

beckermr: They accept lists already.

PR #537, opened 2023-09-01
convert a few patches from the core conda stack

jaimergp:
Checklist

* [X] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/patch_yaml/README.md)).

@beckermr - as discussed in the call, these are the patches I worked on during the flight. I am not sure if we want to keep them in the Python module now though, so let me know what to do with this. At any rate, at least the work is there, worst case we just close.

### DIFF
--- a/recipe/patch_yaml/conda-smithy.yaml
+++ b/recipe/patch_yaml/conda-smithy.yaml
@@ -1,0 +1,25 @@
+# yaml-language-server: $schema=../patch_yaml_model.json
+# # conda moved to calvar from semver and this broke old versions of
+# # conda smithy that do on-the-fly version checks
+# if (
+#     record_name == "conda-smithy"
+#     and (
+#         parse_version(record["version"]) <=
+#         parse_version("3.21.1")
+#     )
+# ):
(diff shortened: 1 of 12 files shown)


recipe/patch_yaml/conda-smithy.yaml
```diff
@@ -0,0 +1,25 @@
...
+#             if len(parts) == 1:
+#                 parts.append("<5a0")
+#             elif len(parts) == 2 and "<" not in parts[1]:
+#                 parts[1] = parts[1] + ",<5a0"
+#             record["depends"][i] = " ".join(parts)
+if:
+  name: conda-smithy
+  version_le: "3.21.1"
+then:
+  - tighten_depends:
```
beckermr, line 23: This patch is not the same. tighten depends won't handle cases where the existing pin has no version spec.

beckermr (changes requested):
Let's get this PR to run and see the output of show_diff.py before making a decision.

PR #541, opened 2023-09-03
REF clean out the rest of the cruft 

beckermr:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,10 @@ build:
   noarch: generic
   number: 0
   script:
-    - pre-commit run -a --show-diff-on-failure black
+    - "yamllint -d '{extends: default, rules: {document-start: disable, line-length: {max: 120}}}' patch_yaml/."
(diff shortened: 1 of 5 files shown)


recipe/meta.yaml
```diff
@@ -11,13 +11,7 @@ build:
...
   number: 0
   script:
-    - "yamllint -d '{extends: default, rules: {document-start: disable, line-length: {max: 120}}}' patch_yaml/."
-    - black --check patch_yaml_utils.py
-    - ruff patch_yaml_utils.py
-    - black --check patch_yaml_model.py
-    - ruff patch_yaml_model.py
-    - black --check test_patch_yaml_utils.py
-    - ruff test_patch_yaml_utils.py
+    - pre-commit run -a --show-diff-on-failure black
```
beckermr, line 14:
```suggestion
    - pre-commit run -a --show-diff-on-failure
```

beckermr:
Diff is zero so merging!

PR #579, opened 2023-10-25
ENH add mpi 1.0 impi metapackage pin for intel mpi

beckermr:
Checklist

* [x] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [x] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.
* [x] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [x] Ran `python show_diff.py` and posted the output as part of the PR.
* [x] Modifications won't affect packages built in the future. 

<!-- Put any other comments or information here --!>

PR #580, opened 2023-10-25
ENH add mpi metapackage to impi-devel

beckermr:
Checklist

xref: https://github.com/conda-forge/mpi-feedstock/issues/11
follows #579


leofang:
I think this should be just done to `impi_rt` (which the -devel package depends on)?

leofang:
Sorry nvm, I take it back, I think #579 should be enough (for the same reason,  the -devel package depends on the _rt package, so it gets the same constraint)

PR #581, opened 2023-10-30
add missing comma in fractal-server constraint

jaimergp:
Checklist

Fixes https://github.com/conda-forge/fractal-server-feedstock/issues/9

### DIFF
--- a/recipe/patch_yaml/fractal-server.yaml
+++ b/recipe/patch_yaml/fractal-server.yaml
@@ -1,0 +1,9 @@
+# yaml-language-server: $schema=../patch_yaml_model.json
+# https://github.com/conda-forge/fractal-server-feedstock/issues/9
+if:
+  name: fractal-server
+  timestamp_lt: 1698663374000  # 2023-10-30
+then:
+  - replace_constrains:
+      old: cloudpickle >=2.2.1<2.3.0
+      new: cloudpickle >=2.2.1,<2.3.0


jaimergp:
```diff
noarch::fractal-server-1.3.9-pyhd8ed1ab_0.conda
noarch::fractal-server-1.3.11-pyhd8ed1ab_0.conda
noarch::fractal-server-1.3.10-pyhd8ed1ab_0.conda
noarch::fractal-server-1.3.12-pyhd8ed1ab_0.conda
-    "cloudpickle >=2.2.1<2.3.0",
+    "cloudpickle >=2.2.1,<2.3.0",


patching repodata:  11%|█         | 1/9 [02:55<23:27, 175.89s/it]

patching repodata:  22%|██▏       | 2/9 [02:57<08:32, 73.15s/it] ================================================================================
================================================================================
linux-armv7l



patching repodata:  33%|███▎      | 3/9 [05:06<09:54, 99.05s/it]================================================================================
================================================================================
linux-aarch64



patching repodata:  44%|████▍     | 4/9 [07:06<08:55, 107.08s/it]================================================================================
================================================================================
linux-ppc64le


patching repodata:  56%|█████▌    | 5/9 [07:38<05:19, 79.88s/it] 
================================================================================
================================================================================
linux-64


patching repodata:  67%|██████▋   | 6/9 [09:17<04:19, 86.62s/it]
================================================================================
================================================================================
osx-arm64



patching repodata:  78%|███████▊  | 7/9 [09:45<02:14, 67.25s/it]================================================================================
================================================================================
win-32


patching repodata:  89%|████████▉ | 8/9 [13:52<02:04, 124.73s/it]
================================================================================
================================================================================
osx-64


patching repodata: 100%|██████████| 9/9 [14:38<00:00, 99.89s/it] 
patching repodata: 100%|██████████| 9/9 [14:38<00:00, 97.57s/it]

================================================================================
================================================================================
win-64

```

tcompa:
> @tcompa can you take a look here as well? This will fix the already published packages.

The diff looks good, thanks.

Side question: do we also need to remove cloudpickle from the `run` block in here (as you did in the feedstock PR), so that it only belongs to the `run_constrained` one? Not a big deal if we keep it in there, though, given it's relatively lightweight.

jaimergp:
> Side question: do we also need to remove cloudpickle from the run block in here (as you did in the feedstock PR), so that it only belongs to the run_constrained one? Not a big deal if we keep it in there, though, given it's relatively lightweight.

We could, but it's not needed for the fix and I prefer having less invasive patches. Considering we already have a fixed build, I'd say we should leave it as is.

tcompa:
> We could, but it's not needed for the fix and I prefer having less invasive patches. Considering we already have a fixed build, I'd say we should leave it as is.

Fine with me

jaimergp:
```diff
patching repodata:   0%|          | 0/9 [00:00<?, ?it/s]
================================================================================
================================================================================
noarch
noarch::fractal-server-1.3.10-pyhd8ed1ab_0.conda
noarch::fractal-server-1.3.12-pyhd8ed1ab_0.conda
noarch::fractal-server-1.3.9-pyhd8ed1ab_0.conda
noarch::fractal-server-1.3.11-pyhd8ed1ab_0.conda
-    "cloudpickle >=2.2.1<2.3.0",
+    "cloudpickle >=2.2.1,<2.3.0",
noarch::mechanize-0.4.9-pyhd8ed1ab_0.conda
-    "python =2.7|>=3.5"
+    "python ==2.7.*|>=3.5"
```

PR #588, opened 2023-11-11
Patch ffmpeg 6.0.0 and 6.0.1 for stricter pinning

hmaarrfk:
It seems like the SO name changed
https://github.com/conda-forge/ffmpeg-feedstock/pull/216
Checklist


hmaarrfk:
i'm not so sure this isneeded anymore, it may have been an error in my grep.

hmaarrfk:
don't

PR #601, opened 2023-11-28
Parametrize the pydantic validation for better error message

hmaarrfk:
Checklist

### DIFF
--- a/recipe/test_patch_yaml_utils.py
+++ b/recipe/test_patch_yaml_utils.py
@@ -529,8 +529,7 @@ def test_schema_up_to_date():
     assert schema_str == schema_on_disk
 
 
-patch_yaml_files = list((Path(__file__).parent / "patch_yaml").glob("*.yaml"))
-patch_yaml_files.sort()
+patch_yaml_files = sorted((Path(__file__).parent / "patch_yaml").glob("*.yaml"))
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


recipe/test_patch_yaml_utils.py
```diff
@@ -529,14 +529,14 @@ def test_schema_up_to_date():
...
-    for document in data.glob("*.yaml"):
-        for patch in yaml.safe_load_all(document.read_text()):
-            try:
-                PatchYaml(**patch)
-            except Exception as e:
-                print(f"\nSchema error in {document.name}: {e}")
-                passed = False
-    assert passed, "schema validation failed!"
+patch_yaml_files = list((Path(__file__).parent / "patch_yaml").glob("*.yaml"))
+patch_yaml_files.sort()
```
jaimergp, line 533:
```suggestion
patch_yaml_files = sorted((Path(__file__).parent / "patch_yaml").glob("*.yaml"))
```

jaimergp:
Is this slower by any chance? We should check. I remember seeing drastic performance improvements in constructor when we switched from pytest parametrization to an inner for loop. That had like thousands of cases, not lower hundreds like here, but just in case, I thought I'd ask.

hmaarrfk:
> Is this slower by any chance

this in my mind is the most worthwhile "test" that we can run and communicate to contributors.

I will check the logs and see if there is any appreciable difference:


This PR:
Start time: `2023-11-28T19:37:34.5421079Z`
End time: `2023-11-28T19:37:34.5653805Z`

under 1 second on 19 mins... i feel like we are ok

jaimergp (approved):

PR #604, opened 2023-12-04
Update pyspectral python lower bound to 3.10

djhoese:
Checklist

```
$ python show_diff.py --use-cache --subdirs noarch
noarch
noarch::pyspectral-0.13.0-pyhd8ed1ab_0.conda
-    "python >=3.7",
+    "python >=3.10",
```


djhoese:
Very strange that pre-commit doesn't catch the extra blank line in the YAML nor does the diff script, but the CI fails because of it.

pnuu (approved):
LGTM

PR #609, opened 2023-12-05
Address issue with more modern builds of libpq

hmaarrfk:
cc: @h-vetinari @akrherz  
Checklist

### DIFF
--- a/recipe/patch_yaml/libpq.yaml
+++ b/recipe/patch_yaml/libpq.yaml
@@ -1,7 +1,8 @@
 # see https://github.com/conda-forge/postgresql-feedstock/issues/178
+# and https://github.com/conda-forge/postgresql-feedstock/issues/191
 if:
   name: libpq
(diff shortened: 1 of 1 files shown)


hmaarrfk:
The packages might as well be called broken:
```
osx-arm64::libpq-15.4-h18bfe24_4.conda
osx-arm64::libpq-16.1-hd435d45_1.conda
osx-arm64::libpq-16.1-hd435d45_2.conda
osx-arm64::libpq-15.4-hd435d45_3.conda
osx-arm64::libpq-16.1-h9aca644_4.conda
-    "openssl >=3.2.0,<4.0a0"
+    "openssl >=3.2.0,<3.2.0a0"
```
means they are uninstallable.

hmaarrfk:
As mentionned in https://github.com/conda-forge/postgresql-feedstock/issues/191 

it seems like only the new version 16 builds are affected, and they might just be outright broken given the constraints.

PR #611, opened 2023-12-06
Add HDF5 1.14.2 -> 1.14.3 ABI compatibility

hmaarrfk:
cc: @conda-forge/hdf5

Checklist


jakirkham:
Thanks Mark! 🙏

Let's give it a go

If something crops up, we can always refine this approach as needed

minrk:
That's a good idea, I'll start propagating that to other packages. I believe #614 fixes the pattern to work for the build string pinners in the meantime.

apmorton:
This migration causes issues with libraries like h5py which check for version mismatch (and probably many others).
https://github.com/conda-forge/h5py-feedstock/issues/122#issuecomment-1849621800

jakirkham:
It is just warning though. Not erroring right?

apmorton:
it is just a warning, yes.

Many projects (like mine) treat warnings as errors in test suites, and so now we must carry around suppressions.

jakirkham:
Thanks for confirming 🙏

Maybe it is worth [asking upstream]( https://github.com/h5py/h5py/issues/new ) if they have better options to suppress this warning?

hmaarrfk:
I don't think we want to create new warnings for users with these migrations.

We could exclude h5py from this migration and rebuild h5py with 1.14.3

hmaarrfk:
I couldn't figure out how to do "all packages except" but maybe we can add a constraint to h5py to restrict the upper bond that way

PR #613, opened 2023-12-07
Add new dependency for openslide 4.0.0 on libdicom

hmaarrfk:
xref: https://github.com/conda-forge/openslide-feedstock/pull/25

cc: @kaczmarj

Checklist

### DIFF
--- a/recipe/patch_yaml/openslide.yaml
+++ b/recipe/patch_yaml/openslide.yaml
@@ -1,0 +1,7 @@
+if:
+  name: openslide
+  version: 4.0.0
+  build_number: 0
+
+then:
+  - add_depends: libdicom >=1.0.5,<1.1.0a0


jakirkham:
Was this dependency also added to the package somewhere?

jakirkham:
Also thanks for looking into this Mark! 🙏

hmaarrfk:
Yes, new dependency in version 4.0.0, but we didn't have a run_export in the package. nor a test in the binary to catch it.

jakirkham:
Ah ok. That context helps. Thanks Mark 🙏

So the steps were:

* Add a `run_export` to `libdicom` ( https://github.com/conda-forge/libdicom-feedstock/pull/5 )
* Patch the existing `openslide` packages

Should we add `libdicom` as a `run` dependency in `openslide` as well (just in case an older `libdicom` package without the `run_export` is used for the build)?

hmaarrfk:
> Should we add libdicom as a run dependency in openslide as well (just in case an older libdicom package without the run_export is used for the build)?

no a test is being added instead.

jakirkham:
Ok thanks! Updated the summary above. Does that look better

hmaarrfk:
yes.

PR #614, opened 2023-12-08
capture build string variations in hdf5 1.14.3 patch

minrk:
Update rule in #611 to capture dependencies with build string pinning, which would be anything that build against hdf5 with mpi, e.g. petsc.

Checklist

[showdiff output](https://gist.github.com/minrk/9d0e77781155d14bb5b9095734bfabb3)

### DIFF
--- a/recipe/patch_yaml/hdf5.yaml
+++ b/recipe/patch_yaml/hdf5.yaml
@@ -2,8 +2,8 @@
 # hdf5 1.14.2 and 1.14.3 have the same ABI
 # https://gamma.hdfgroup.org/ftp/pub/outgoing/hdf5/prelim_reports/hdf5-1.14.2-vs-hdf5-1.14.3-1-interface_compatibility_report.html
 if:
-  has_depends: hdf5 >=1.14.2,<1.14.3.0a0
-  timestamp_lt: 1701868616000
+  has_depends: hdf5 >=1.14.2,<1.14.3.0a0?( *)
+  timestamp_lt: 1702066879000
 
 then:
   - loosen_depends:


recipe/patch_yaml/hdf5.yaml
```diff
@@ -2,8 +2,8 @@
 # hdf5 1.14.2 and 1.14.3 have the same ABI
 # https://gamma.hdfgroup.org/ftp/pub/outgoing/hdf5/prelim_reports/hdf5-1.14.2-vs-hdf5-1.14.3-1-interface_compatibility_report.html
 if:
-  has_depends: hdf5 >=1.14.2,<1.14.3.0a0
-  timestamp_lt: 1701868616000
+  has_depends: hdf5 >=1.14.2,<1.14.3.0a0?( *)
+  timestamp_lt: 1702066879000
```
jakirkham, line 6: Could you please share a bit about the timestamp change? It appears the new timestamp is more recent than the previous one

jakirkham (commented):
Thanks Min! 🙏

Had one question above

recipe/patch_yaml/hdf5.yaml
jakirkham: Wonder if we should just rebuild the `hdf5` packages to include a better `run_exports` line

While in theory dropping the timestamp condition would mean anything would be patched. Really it would mean we patch the repodata based on whenever this feedstock last built a package with a new patch. Though when that happens is kind of arbitrary

hmaarrfk: I think this is really out of scope for now.

We should really just answer this question in my mind:
https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/3368

For now this PR looks like a step in the right direction. the timestamp still exists so itisn't so wide reaching, but there is still a risk of systems being built with 1.14.2 instead of 1.14.3 that would carry the 'x.x.x' limitations

hmaarrfk:
Thank you.

PR #619, opened 2023-12-14
Require menuinst>=2 in conda-build 3.28.0 and 3.28.1

jaimergp:
Checklist

- https://github.com/AnacondaRecipes/repodata-hotfixes/pull/218

### DIFF
--- a/recipe/patch_yaml/conda-build.yaml
+++ b/recipe/patch_yaml/conda-build.yaml
@@ -79,3 +79,13 @@ then:
   - tighten_depends:
       name: conda
       upper_bound: 23.9.0
+---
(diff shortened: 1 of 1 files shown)


xhochy:
@jaimergp Is this ready? I'm missing a diff here as an indicator, otherwise would be fine to merge.

jaimergp:
Same, 3.28.0 is not patched. Is it because it's marked as broken? Anyway, ready for review and merge @xhochy

jaimergp:
Ah, 3.28.1 is marked as broken now too, so maybe we don't need this after all. 3.28.2 should pin this adequately.

hmaarrfk:
ok to close?

PR #628, opened 2024-01-02
tighten fmt in libmamba

jaimergp:
Checklist

* [X] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [ ] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.
* [ ] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [ ] Ran `python show_diff.py` and posted the output as part of the PR.
* [X] Modifications won't affect packages built in the future. 

`fmt 10.2.0` breaks libmamba on Windows: https://github.com/conda-forge/fmt-feedstock/pull/39

PR #629, opened 2024-01-03
Revert "tighten fmt in libmamba"

h-vetinari:
fmt 10.2 accidentally dropped a symbol that lead to widespread breakage on windows, including all of our windows builds (because it affected mamba itself). After #628 stopped the bleeding, @saraedum went digging and rootcaused+fixed this upstream in https://github.com/fmtlib/fmt/pull/3786, which already made it to conda-forge: https://github.com/conda-forge/fmt-feedstock/pull/42

This means we should be able to drop the pin again, as over time, more and more packages will now build with `fmt 10.2` and thus get a `fmt >=10.2,<11.0a0` run-export, that would be incompatible with the patched constraints for mamba from #628.

An alternative would be to just rebuild mamba once for 10.2, and leave the older builds constrained to 10.1, though I don't really see the need for this if the ABI is fixed.

CC @jaimergp

### DIFF
--- a/recipe/patch_yaml/mamba.yaml
+++ b/recipe/patch_yaml/mamba.yaml
@@ -1,4 +1,3 @@
-# yaml-language-server: $schema=../patch_yaml_model.json
 # from this code
 # if (
 #     subdir in ["linux-64", "linux-aarch64", "linux-ppc64le"]
@@ -142,13 +141,3 @@ then:
   - replace_depends:
       old: conda >=23.9,<24
       new: conda >=23.9,<23.11.0
----
-# fmt 10.2.0 breaks libmamba on Windows
-# https://github.com/conda-forge/fmt-feedstock/pull/39
-if:
-  name: libmamba
-  timestamp_lt: 1704218331614
-then:
-  - tighten_depends:
-      name: fmt
-      upper_bound: "10.2.0"


hmaarrfk:
I believe the offending build was marked as broken in https://github.com/conda-forge/admin-requests/pull/902 is that correct?

SylvainCorlay:
> I believe the offending build was marked as broken in https://github.com/conda-forge/admin-requests/pull/902 is that correct?

Yes, it seems so. Do we need to wait for CDNs to synchronize before merging ?

jakirkham:
The patches were regenerated in [this CI job]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=852421&view=results ). May take a bit for the CDN to pick that up

Though got the sense this change might be less urgent. This based on the fact Axel asked Jaime to review (who is likely done for the day) and presented options for discussion. Also got the sense that other fixes were already in place (this is just relaxing some of them)

Please let me know if I'm missing something though. Haven't caught up on all the related threads yet

bdice (approved):

h-vetinari:
> The patches were regenerated in [this CI job](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=852421&view=results). May take a bit for the CDN to pick that up

The patches from #628 were picked up quickly yesterday, which is what "unbroke" our windows pipelines. Now that a bugfix was identified and backported, we are able to undo that workaround (if we want).

> Though got the sense this change might be less urgent. This based on the fact Axel asked Jaime to review (who is likely done for the day) and presented options for discussion. Also got the sense that other fixes were already in place (this is just relaxing some of them)

💯

jaimergp (approved):
LGTM

ryanvolz:
Sorry for the ping, but I'd love to see this merged to resolve a conflict I'm seeing between `libmamba` and recent builds of a few packages that depend on `fmt`. (This is in the context of updating an installer built with conda constructor, where I haven't transitioned yet to having the installer create extra environments besides the base env.)

PR #637, opened 2024-01-18
Allow 0.190 for elfutils=0.18*

xhochy:
Checklist


xhochy:
DIff https://gist.github.com/xhochy/014629c5313a7909a31d59e3b7587181

PR #642, opened 2024-01-24
Add a patch to pin openblas <0.3.26 for older scipy builds

hmaarrfk:
xref: https://github.com/conda-forge/scipy-feedstock/issues/266#issuecomment-1907148347

cc: @h-vetinari 

diff: https://gist.github.com/hmaarrfk/b69c7844166efc7cd61f98ae2a4548d5

Checklist

### DIFF
--- a/recipe/patch_yaml/scipy.yaml
+++ b/recipe/patch_yaml/scipy.yaml
@@ -1,0 +1,26 @@
+# 2024/01 -- hmaarrfk
+# scipy version 1.12.0 build 2 was patched
+# for compatibility with libopenblas 0.3.26
+# However older version are incompatible
+# https://github.com/conda-forge/scipy-feedstock/issues/266
+if:
+  name: scipy
+  version: 1.12.0
+  build_number_lt: 2
+
+then:
+  - add_constrains: libopenblas <0.3.26
+
+---
+
+# 2024/01 -- hmaarrfk
+# scipy version 1.12.0 build 2 was patched
+# for compatibility with libopenblas 0.3.26
+# However older version are incompatible
+# https://github.com/conda-forge/scipy-feedstock/issues/266
+if:
+  name: scipy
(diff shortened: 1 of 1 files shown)


h-vetinari:
I think the approach is correct, but this is touching _a lot_ of builds (even though the addition of a constraint for a package that didn't exist at the time these builds were published should be pretty harmless). Not sure what the best way to handle this is. CC @conda-forge/core.

recipe/patch_yaml/scipy.yaml
```diff
@@ -0,0 +1,26 @@
+# 2024/01 -- hmaarrfk
+# scipy version 1.20.0 build 2 was patched
```
h-vetinari, line 2:
```suggestion
# scipy version 1.12.0 build 2 was patched
```

recipe/patch_yaml/scipy.yaml
```diff
@@ -0,0 +1,26 @@
...
+  version: 1.12.0
+  build_number_lt: 2
+
+then:
+  - add_constrains: libopenblas <0.3.26
+
+---
+
+# 2024/01 -- hmaarrfk
+# scipy version 1.20.0 build 2 was patched
```
h-vetinari, line 17:
```suggestion
# scipy version 1.12.0 build 2 was patched
```

h-vetinari (approved):
LGTM aside from some comment nits.

recipe/patch_yaml/scipy.yaml
hmaarrfk: thx!

h-vetinari:
Any comments? This is hitting scipy installs quite widely so I think we should do it (e.g. someone who's not able to update to 1.12, which is the only version that's fixed ATM)

beckermr:
merge away!

PR #657, opened 2024-02-08
Tighten conda-build dep in boa

jaimergp:
Checklist

Addresses incompatibility disclosed at https://github.com/mamba-org/boa/issues/392


jaimergp:
```diff
noarch
noarch::boa-0.16.0-pyhd8ed1ab_0.conda
noarch::boa-0.16.0-pyhd8ed1ab_1.conda
-    "conda-build >=3.24",
+    "conda-build >=3.24,<24.1.0a0",
```

Note that even if this patch contains 0.15.x, that release series is marked as broken, so it's not patched. Only added them for completeness.

PR #685, opened 2024-03-19
Avoid conda 24.3.0 breaking conda-build

jaimergp:
Checklist

`conda` 24.3.0 deprecated some stuff that conda-build still used (see https://github.com/conda/conda-build/pull/5221).

### DIFF
--- a/recipe/patch_yaml/conda-build.yaml
+++ b/recipe/patch_yaml/conda-build.yaml
@@ -86,6 +86,7 @@ if:
   name: conda-build
   version_gt: 3.21.8
   version_lt: 24.3.0
+  timestamp_le: 1710834612324  # 2024-03-19
 then:
   - tighten_depends:
       name: conda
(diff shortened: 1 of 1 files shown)


h-vetinari (approved):

xhochy (approved):

h-vetinari:
If we end up updating the run-constraint in https://github.com/conda-forge/conda-feedstock/pull/232, then this repodata patch is actually not strictly speaking necessary (but still doesn't hurt). Inversely, now that we have the repodata-patch, we don't categorically _need_ the run-constraint, but it would IMO be the cleaner (because more explicit) way to do it.

jezdez:
Thanks all!

mbargull:
BTW, upstream is preparing some CI checks such that we should catch these things much earlier: https://github.com/conda/conda-build/pull/5244 .

PR #687, opened 2024-03-19
imod 0.15.2 and 0.15.3; Pin xarray to higher version: v2023.08

JoerivanEngelen:
Checklist

We introduced a call to ``xarray.Coordinates``, which was introduced in xarray v2023.08.0, so we have to pin these versions to a higher level. I had some issues running ``show_diff.py`` on my machine, but I read the CI will also run it, so will post the output here when it is done.

UPDATE
---------
The CI prints:

```
noarch
noarch::imod-0.15.2-pyhd8ed1ab_0.conda
noarch::imod-0.15.3-pyhd8ed1ab_0.conda
-    "xarray >=2023.04.0",
+    "xarray >=2023.08.0",
```

So it does what it needs to do


JoerivanEngelen:
@conda-forge/core Could you review this PR?

PR #690, opened 2024-03-26
Add patch for j2cli to be incompatible with python 3.12

hmaarrfk:
Checklist


hmaarrfk:
```
noarch
noarch::j2cli-0.3.10-py_0.tar.bz2
-    "python",
+    "python <3.12.0a0",
```

PR #717, opened 2024-04-23
BUG make sure skyproj uses mpl base >=3.6

beckermr:
Checklist


recipe/patch_yaml/skyproj.yaml
```diff
@@ -0,0 +1,8 @@
+if:
+  name: skyproj
+  version_in: [1.2.3, 1.2.3]
```
beckermr, line 3:
```suggestion
  version_in: [1.2.2, 1.2.3]
```

PR #719, opened 2024-04-29
unmask build variable in repodata patch templating

croth1:
Checklist

xref: https://github.com/conda-forge/conda-forge.github.io/discussions/2160 
cc @jaimergp

I am still not sure whether there are guarantees that two outputs generated by the same build process will always share the same build hash (basically what exactly goes into the hash function).


beckermr (approved):

croth1:
uhum - such a simple change that I didn't bother checking the test - gimme a minute 😬

beckermr:
thank you!

PR #727, opened 2024-05-12
pyarrow: Constrain pandas<3 if pyarrow<15

IzerOnadimQC:
Checklist

Pyarrow versions less than 15 result in an `AttributeError` when used with nightly builds of Pandas (>=3) due to breaking API changes in Pandas. Therefore, constrain the Pandas version to less than 3 if the Pyarrow version is less than 15.

### DIFF
--- a/recipe/patch_yaml/pyarrow.yaml
+++ b/recipe/patch_yaml/pyarrow.yaml
@@ -29,3 +29,17 @@ then:
       name: numpy
       max_pin: x
       upper_bound: "1.20"
+---
+# Prior to Pyarrow 15.0.0 the pandas_compat code made use of pandas.core.internals.DatetimeTZBlock,
+# which will no longer be exposed as of Pandas 3. This means that converting a Table to a Dataframe
+# using Pyarrow<15.0.0 and Pandas>=3 will result in an AttributeError. Therefore, we constrain the
+# Pandas version to less than 3 if the Pyarrow version is less than 15.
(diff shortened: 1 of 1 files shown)


IzerOnadimQC:
Some questions:

- @xhochy Is this the correct way to add a constraint on an optional dependency?
  - Also, this leads to a pretty enormous repo data change. Is this an issue? I could probably make the diff smaller by only applying the change to more recent versions rather than all of them.
- @jorisvandenbossche do you agree that this constraint makes sense given: https://github.com/apache/arrow/issues/38341?

jorisvandenbossche:
I would personally prefer to _not_ add this version constraint, but rather ensure pandas does not yet remove that API (I added a comment about that in https://github.com/pandas-dev/pandas/pull/58467#issuecomment-2110042930).

Of course that will still give you a warning when having this combination of pandas 3.0 with older pyarrow, which is maybe also something you want to avoid, but I would say that is then up to the user to decide if they want to ensure to use a more recent pyarrow version for that or not (the warning will also only be shown for specific cases, i.e. when having timezone aware timestamps)

xhochy:
> In that case, I will hold off on this PR until a decision is made regarding which version of Pandas will remove the DatetimeTZBlock. Then I'll add a constraint for that version.

Marked as draft until then.

Otherwise, this PR looks good.

hmaarrfk:
The missing context for was that pandas 3 is still unreleased as of May 19, 2024.
https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html

As such, this can wait for a resolution upstream.

PR #739, opened 2024-05-23
loosen zlib-pin to major level (1/N)

h-vetinari:
Restricted to monthly cadence after discussion in https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/5959

Previously:
> This is a draft because it impacts a large number of packages, almost 100k 😑
> 
> That's even though we're "only" patching 2.5 years worth of packages, since [conda-forge/zlib-feedstock@a97f1e6](https://github.com/conda-forge/zlib-feedstock/commit/a97f1e6e3b2288f1b225f9976b2d40cf7231ddaa) changed the run-export to `libzlib`.
> 
> Here's a [gist](https://gist.github.com/h-vetinari/dc5e9d3aa76408ad10cb493dd5b28a6a) with the affected packages.

PR #740, opened 2024-05-28
loosen zlib-pin to major level (2/N)

h-vetinari:
Follow-up to #739. Will run the diff once the CDN has synched after that PR.

### Result of `python show_diff.py`

<details>

```diff
================================================================================
================================================================================
linux-armv7l
================================================================================
================================================================================
win-32
================================================================================
================================================================================
osx-arm64
osx-arm64::pyarrow-tests-14.0.2-py39hb31e23e_18_cpu.conda
osx-arm64::libadios2-2.9.2-mpi_mpich_h0aa738d_2.conda
osx-arm64::libarrow-14.0.2-ha642a4c_19_cpu.conda
osx-arm64::heyoka-llvm-15-4.0.3-h510fb61_2.conda
osx-arm64::ndcctools-7.9.1-py39hec72b60_0.conda
osx-arm64::libadios2-2.10.0-mpi_openmpi_h135fdc6_1.conda
osx-arm64::lxml-5.1.1-py312h867102f_0.conda
osx-arm64::grpcio-1.62.2-py39h047a24b_0.conda
osx-arm64::tiledb-2.22.0-hb090472_2.conda
osx-arm64::libadios2-2.10.0-mpi_openmpi_h2343a08_2.conda
osx-arm64::nest-simulator-3.6-py312h6157364_5.conda
osx-arm64::pyarrow-14.0.2-py312h9cc9a04_18_cpu.conda
osx-arm64::libgdal-arrow-parquet-3.7.3-haddc5de_22.conda
osx-arm64::heyoka-4.0.3-h4dff030_2.conda
osx-arm64::lammps-2024.02.07-cpu_py38_hb5f05e6_mpi_mpich_0.conda
osx-arm64::folly-2024.04.01.00-h25af3d9_0.conda
osx-arm64::heyoka-4.0.3-h0f46c6c_0.conda
osx-arm64::util-linux-2.39.4-py311h5768e45_0.conda
osx-arm64::poppler-24.04.0-h42742f0_0.conda
osx-arm64::libadios2-2.9.2-mpi_openmpi_h6eedf57_2.conda
osx-arm64::libadios2-2.10.0-mpi_openmpi_h6eedf57_2.conda
osx-arm64::sqlite-3.45.3-hf2abe2d_0.conda
osx-arm64::gazebo-11.14.0-origname1234567_211.conda
osx-arm64::heyoka-4.0.3-he1ce634_1.conda
osx-arm64::libgdal-3.7.3-h5dd8421_22.conda
[...]

PR #741, opened 2024-05-29
loosen zlib-pin to major level (3/N)

h-vetinari:
See #739


h-vetinari:
Seems the CDN hasn't updated yet - `show_diff.py` is still picking up the changes from #740

beckermr:
There is a CDN outage today.

PR #755, opened 2024-06-03
Fix mismatched dependencies for arviz 0.13.0

OriolAbril:
Checklist

### DIFF
--- a/recipe/patch_yaml/arviz.yaml
+++ b/recipe/patch_yaml/arviz.yaml
@@ -9,7 +9,7 @@ then:
 # ArviZ ==0.13.0 has mismatched dependencies with respecto to pypi
 if:
   name: arviz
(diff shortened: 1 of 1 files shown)


recipe/patch_yaml/arviz.yaml
```diff
@@ -5,3 +5,12 @@ if:
   version_le: "0.17.1"
 then:
   - add_depends: scipy <1.13
+
```
OriolAbril, line 8:
```suggestion

---
```

OriolAbril:
went with https://github.com/conda-forge/admin-requests/pull/1010 instead

maresb:
Now doing https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/774/files

PR #776, opened 2024-06-12
Mark pint 0.24 build 0 requires Python 3.10 rather than 3.9

jan-janssen:
The pint package is released as `noarch` but dropped support for python 3.9 as reported in https://github.com/conda-forge/pint-feedstock/issues/65 . A new build was released https://github.com/conda-forge/pint-feedstock/pull/66 but on Python 3.9 conda still tries to install the previous build, so it should be marked as broken.

Based on the feedback in https://github.com/conda-forge/admin-requests/pull/1017

Checklist

### DIFF
--- a/recipe/patch_yaml/pint.yaml
+++ b/recipe/patch_yaml/pint.yaml
@@ -17,7 +17,7 @@ then:
 # pint 0.24 requires Python 3.10 and does not run on older versions
 if:
   name: pint
-  version: 0.24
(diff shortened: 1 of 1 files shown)


jan-janssen:
```
noarch
noarch::pint-0.24-pyhd8ed1ab_0.conda
-    "python >=3.9",
+    "python >=3.10",
```

PR #783, opened 2024-06-26
urllib3 2.2.2 not compatible with python 3.7

AaronOpfer:
See https://github.com/conda-forge/urllib3-feedstock/issues/84 and https://github.com/conda-forge/urllib3-feedstock/pull/85

Checklist

```
linux-armv7l
win-32
osx-arm64
linux-ppc64le
linux-aarch64
noarch
noarch::urllib3-2.2.2-pyhd8ed1ab_0.conda
-    "python >=3.7"
+    "python >=3.8"
win-64
osx-64
linux-64
```

### DIFF
--- a/recipe/patch_yaml/urllib3.yaml
+++ b/recipe/patch_yaml/urllib3.yaml
@@ -1,0 +1,10 @@
+# https://github.com/conda-forge/urllib3-feedstock/issues/84
+# urllib3 2.2.2 is not compatible with Python 3.7
+if:
+  name: urllib3
+  version: 2.2.2
+  timestamp_lt: 1719424349134  # 2024/06/26 12:52 GMT
+then:
+  - replace_depends:
+      old: python >=3.7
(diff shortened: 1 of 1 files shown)


AaronOpfer:
@ocefpaf I think this is what you had in mind?

recipe/patch_yaml/urllib3.yaml
```diff
@@ -0,0 +1,10 @@
+# https://github.com/conda-forge/urllib3-feedstock/issues/84
+# urllib3 2.2.2 is not compatible with Python 3.7
+if:
+  name: urllib3
+  version: 2.2.2
```
ocefpaf, line 5: It doesn't hurt to be a bit more specific about the artifact we are patching.

```suggestion
  version: 2.2.2
  build_number_in: [0]
```

jakirkham:
Thank you both! 🙏

PR #675, opened 2024-02-27
Add conda environment and tips for testing patches locally

jdblischak:
Checklist

* [x] Ran `pre-commit run -a` and ensured all files pass the linting checks.

---

I added a conda environment that includes all the dependencies required to run `pre-commit` and `show_diff.py` (and no more, since it's not needed to run `gen_patch_json.py` locally). I also added some tips to the README that I would have found helpful when preparing repodata patches in the past.

Closes #493

PR #798, opened 2024-07-12
Add pixi

wolfv:
Checklist

### DIFF
--- a/recipe/pixi.toml
+++ b/recipe/pixi.toml
@@ -1,0 +1,18 @@
+[project]
+name = "repodata-patches"
+version = "0.1.0"
+description = "Add a short description here"
+authors = ["Wolf Vollprecht <w.vollprecht@gmail.com>"]
+channels = ["conda-forge", "nodefaults"]
+platforms = ["osx-arm64"]
+
+[tasks]
+show-diff = "python ./show_diff.py"
+
+[dependencies]
+annotated-types = "*"
+conda-build = "*"
+license-expression = "*"
+pre-commit = "*"
+pydantic = "*"
+python = "*"
--- a/recipe/pixi.toml
+++ b/recipe/pixi.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 description = "Add a short description here"
 authors = ["Wolf Vollprecht <w.vollprecht@gmail.com>"]
 channels = ["conda-forge", "nodefaults"]
-platforms = ["osx-arm64"]
+platforms = ["osx-arm64", "linux-64", "win-64"]
 
 [tasks]
 show-diff = "python ./show_diff.py"
--- a/recipe/.gitignore
+++ b/recipe/.gitignore
@@ -3,3 +3,6 @@ tmp
 __pycache__
 **/.pytest_cache/
 **/.ruff_cache/
+# pixi environments
+.pixi
+*.egg-info
--- a/recipe/pixi.lock
+++ b/recipe/pixi.lock
@@ -5,6 +5,130 @@ environments:
     - url: https://conda.anaconda.org/conda-forge/
     - url: https://conda.anaconda.org/nodefaults/
     packages:
+      linux-64:
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/noarch/annotated-types-0.7.0-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/archspec-0.2.3-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/attrs-23.2.0-pyh71513ae_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/beautifulsoup4-4.12.3-pyha770c72_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/boltons-24.0.0-pyhd8ed1ab_0.conda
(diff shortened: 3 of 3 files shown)


hmaarrfk:
is a pixi lock file for a "small project" such as this really 5k lines long?

wolfv:
It will become a little shorter soon (we remove some duplicate data in a upcoming PR). But yeah, since we need the repodata records it is rather long.

bollwyvl:
I think the approach from `staged-recipes` of _not_ checking in the `pixi.lock` is going to be the right play here. On account of that, the bottom dependency versions should be tightened a fair amount and kept in sync with the `meta.yaml` so that migrations trickle down into `pixi.toml` and trigger a re-lock locally as needed.

I could definitely see a useful `pixi r pr` top-level command that:
- `pixi r lint`
  - validated all patches
- `pixi r diff`
  - ran `show_diff.py` 
    - wrote it out to `$(git rev-parse --abbrev-ref HEAD).diff.md` for easy of copy-pasting and review

jdblischak (commented):
Could you please also add some documentation to `recipe/README.md` to alert contributors that the `pixi` configuration exists and how they can use it to test their patches? I'm imagining something similar to #675 where I added a conda env file

recipe/pixi.toml
```diff
@@ -0,0 +1,18 @@
...
+description = "Add a short description here"
+authors = ["Wolf Vollprecht <w.vollprecht@gmail.com>"]
+channels = ["conda-forge", "nodefaults"]
+platforms = ["osx-arm64", "linux-64", "win-64"]
+
+[tasks]
+show-diff = "python ./show_diff.py"
+
+[dependencies]
+annotated-types = "*"
```
bollwyvl, line 13: Already a dependency of `pydantic`...

recipe/pixi.toml
```diff
@@ -0,0 +1,18 @@
...
+conda-build = "*"
+license-expression = "*"
+pre-commit = "*"
+pydantic = "*"
```
bollwyvl, line 17: I... don't believe this can be `*`... I'd wager `2.*` _might_ be safe, but there are a lot of pins out there...

recipe/pixi.toml
```diff
@@ -0,0 +1,18 @@
+[project]
+name = "repodata-patches"
+version = "0.1.0"
+description = "Add a short description here"
+authors = ["Wolf Vollprecht <w.vollprecht@gmail.com>"]
+channels = ["conda-forge", "nodefaults"]
+platforms = ["osx-arm64", "linux-64", "win-64"]
+
+[tasks]
+show-diff = "python ./show_diff.py"
```
bollwyvl, line 10: as this is:
- _very_ expensive, I recommend adding a `description` noting that
- _very_verbose, probably also a task which writes it out to a `> {git branch}.txt`

recipe/pixi.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+annotated-types = "*"
+conda-build = "*"
+license-expression = "*"
+pre-commit = "*"
```
bollwyvl, line 16: `pre-commit` isn't going to fire unless inside a `shell`... having a `task.lint` that runs all the linters against all the files would likely be helpful.

bollwyvl:
I'm not convinced checking in the `pixi.lock` is a net win for this particular project...

recipe/pixi.toml
ruben-arts: I added the descriptor. I didn't commit the branch name addition as it would complicate the readme and .gitignore.

recipe/README.md
```diff
@@ -170,15 +194,14 @@ then:
 
 > [!WARNING]
 > The condition `timestamp_lt` is required to prevent your patch from modifying
-> any packages built in the future. Don't forget to calculate it with `python -c
-> "import time; print(f'{time.time():.0f}000')"` and include it in the `if:`
```
jdblischak, line 174: Why hide the Python code from the user? Why not just append something like "or run `pixi run timestamp`"?

recipe/README.md
```diff
@@ -170,15 +194,14 @@ then:
...
+> any packages built in the future. Don't forget to calculate it with `pixi run timestamp`
+> and include it in the `if:` section of your patch.
 
 ## Testing New Patches using `show_diff.py`
 
 > [!TIP]
-> You can install a development environment for testing your repodata patch
-> using the environment file `dev-env-for-patches.yaml`
+> You can use `pixi` to develop and test your patches locally.
+> Install the environment with `pixi install`.
```
jdblischak, line 204: Why must `pixi` be presented as the only option? I suggest the below:

> You can install a development environment for testing your repodata patch
> using the environment file `dev-env-for-patches.yaml`.
> Alternatively, you can use `pixi` to develop and test your patches locally.
> Install the environment with `pixi install`.

recipe/pixi.toml
```diff
@@ -0,0 +1,26 @@
...
+[tasks.diff-cached]
+depends-on = [{task = "show-diff", args = ["--use-cache"]}]
+
+[tasks.lint]
+cmd = "pre-commit run --all-files"
+description = "Run pre-commit hooks to lint the code."
+
+[tasks.timestamp]
+cmd = "python -c \"import time; print(f'{time.time():.0f}000')\""
+description = "Generate a timestamp in milliseconds since epoch."
```
h-vetinari, line 20: If we make this a task, it should be timezone-aware. Our infra operates on UTC timestamps, but running the code above generates timestamps in the local timezone of the user. We've been getting away with that inaccuracy so far, but now that we're providing an "official" wrapper, it should be done properly.

h-vetinari: As a concrete suggestion how to achieve this
```suggestion
[tasks.timestamp]
cmd = "python -c \"import pandas; print(f'{pandas.Timestamp.now(tz=\"UTC\").timestamp():.0f}000')\""
description = "Generate a timestamp in milliseconds since epoch."
```

ruben-arts: Thank you! I was not aware of this.

PR #805, opened 2024-07-24
fix: show all diffs either way

beckermr:
Checklist

This PR fixes a bug where diffs that involve a package not in the current index do not show. It is a weird edge case that we basically never encounter in practice except when I did.


beckermr:
pre-commit.ci autofix

PR #811, opened 2024-08-01
fix: add upper bound for smithy due to conda-build deprecations

beckermr:
Checklist

### DIFF
--- a/recipe/patch_yaml/conda-smithy.yaml
+++ b/recipe/patch_yaml/conda-smithy.yaml
@@ -49,4 +49,4 @@ if:
 then:
   - tighten_depends:
       name: conda-build
-      upper_bound: 24.7.1
+      upper_bound: 24.7.0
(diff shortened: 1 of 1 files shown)


h-vetinari (approved):

recipe/patch_yaml/conda-smithy.yaml
```diff
@@ -42,3 +42,11 @@ then:
...
       name: conda-build
       upper_bound: 24.5.0
+---
+if:
+  name: conda-smithy
+  version_le: 3.37.1
+then:
+  - tighten_depends:
+      name: conda-build
+      upper_bound: 24.7.1
```
h-vetinari, line 52: as a minor nit, we could make it 24.7.0 (even though that this is probably never going to be released in conda-forge, it would break in the same way)
```suggestion
      upper_bound: 24.7.0
```

beckermr: Good idea. Fixed.

jakirkham (approved):

PR #818, opened 2024-08-05
Constrain pydantic version

ijpulidos:
Checklist


ijpulidos:
@conda-forge-admin Should be ready for review! Thanks.

hmaarrfk:
is there a new version of feflow?

ijpulidos:
@hmaarrfk No, just 0.1.0 for now. But there will be a new one coming soon (patch release supporting pydantic 2).

ijpulidos:
@hmaarrfk Oh do you mean that this could have been done in the `feflow` feedstock itself? Maybe with a different build number, right?

hmaarrfk:
I guess we typically like to patch once the underlying problem has been resolved in the feedstock itself.

hmaarrfk:
so either a new build number with the constraint (that will be removed "momentarily") or a new version that makes the constraint obsolete.

ijpulidos:
@hmaarrfk understood, let me do that!

hmaarrfk:
you can also backport the "small patch"

ijpulidos:
@hmaarrfk The feedstock should be fixed now https://github.com/conda-forge/feflow-feedstock/pull/1 is that enough to unblock this? Please let me know. Thank you for your support!

PR #823, opened 2024-08-15
Fix semver dependency in grayskull

maresb:
Checklist

* [X] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.

### DIFF
--- a/recipe/patch_yaml/grayskull.yaml
+++ b/recipe/patch_yaml/grayskull.yaml
@@ -1,0 +1,35 @@
+# Accidentally unconstrained semver dependency in 0th build, fixed in
+# <https://github.com/conda-forge/grayskull-feedstock/pull/74>
+if:
+  name: grayskull
+  version_eq: "2.3.0"
+  build_number_eq: 0
+  timestamp_lt: 1723758527000
+then:
+  - replace_depends:
+      old: semver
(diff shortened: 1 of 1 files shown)


maresb:
```diff
noarch
noarch::grayskull-2.3.1-pyhd8ed1ab_0.conda
noarch::grayskull-2.4.1-pyhd8ed1ab_0.conda
noarch::grayskull-2.5.2-pyhd8ed1ab_0.conda
noarch::grayskull-2.5.0-pyhd8ed1ab_0.conda
noarch::grayskull-2.6.0-pyhd8ed1ab_0.conda
noarch::grayskull-2.4.0-pyhd8ed1ab_0.conda
noarch::grayskull-2.5.3-pyhd8ed1ab_0.conda
noarch::grayskull-2.5.1-pyhd8ed1ab_0.conda
-    "semver >=3.0.0,<4.0.0",
+    "semver >=3.0.0,<3.1.dev0",
noarch::grayskull-2.3.0-pyhd8ed1ab_0.conda
-    "semver",
+    "semver >=2.13.0,<2.14.dev0",
```

maresb:
CC @BastianZim

Note that the latest semver is v3.0.2, so the difference between the upper pins `<4.0.0` and `<3.1.dev0` does not yet exist, so this is just to prevent problems in the future.

PR #839, opened 2024-09-04
anndata 0.10.9 is incompatible with mudata<0.3.1

jdblischak:
Checklist

* [x] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [ ] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.
* [x] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [x] Ran `python show_diff.py` and posted the output as part of the PR.
* [x] Modifications won't affect packages built in the future. 

---

The recently released anndata 0.10.9 is incompatible with mudata <0.3.1. mudata 0.3.1 was released immediately after 0.3.0 in order to fix this incompatibility.

cc: @conda-forge/mudata, @conda-forge/anndata 
xref: https://github.com/scverse/mudata/issues/77, https://github.com/conda-forge/mudata-feedstock/pull/14, https://github.com/conda-forge/mudata-feedstock/pull/15

Demonstrating the problem:

```sh
# 0.2.0
mamba create --yes -n test-mudata-0.2.0 \
  -c conda-forge --override-channels \
  anndata=0.10.9 mudata=0.2.0
mamba activate test-mudata-0.2.0
python -c "import mudata"
## ImportError: cannot import name 'AlignedViewMixin' from 'anndata._core.aligned_mapping' (/home/wsl/mambaforge/envs/test-mudata-0.2.0/lib/python3.12/site-packages/anndata/_core/aligned_mapping.py). Did you mean: 'AlignedView'?
mamba deactivate

# 0.2.4
mamba create --yes -n test-mudata-0.2.4 \
  -c conda-forge --override-channels \
  anndata=0.10.9 mudata=0.2.4
mamba activate test-mudata-0.2.4
python -c "import mudata"
## ImportError: cannot import name 'AlignedViewMixin' from 'anndata._core.aligned_mapping' (/home/wsl/mambaforge/envs/test-mudata-0.2.4/lib/python3.12/site-packages/anndata/_core/aligned_mapping.py). Did you mean: 'AlignedView'?
mamba deactivate

# 0.3.0
mamba create --yes -n test-mudata-0.3.0 \
  -c conda-forge --override-channels \
  anndata=0.10.9 mudata=0.3.0
mamba activate test-mudata-0.3.0
python -c "import mudata"
[...]

PR #840, opened 2024-09-05
Do not include build_number for mudata patch

jdblischak:
Checklist

---

This is a minor follow-up to my recent mudata patch in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/839.

Right after it was merged, I noticed that I had accidentally included `build_number: 0`, which would result in any binaries with other build numbers being excluded from the patch. Now, just by luck, all the binaries with version <= 0.3.1 happened to also have build number 0. Therefore this PR will have no effect on the repodata. Its purpose is purely to clarify the intent of the original patch (and to avoid the potential for future readers of the patch, especially future me, from thinking that this mistake might have caused a problem).

cc: @xhochy 

```sh
python show_diff.py --subdirs noarch
noarch
```

### DIFF
--- a/recipe/patch_yaml/mudata.yaml
+++ b/recipe/patch_yaml/mudata.yaml
@@ -17,7 +17,6 @@ then:
 if:
   name: mudata
   version_lt: "0.3.1"
-  build_number: 0
   timestamp_lt: 1725468863000
 then:
   - tighten_depends:


jdblischak:
[CI](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1020064&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3) confirmed that this clarification PR does not affect the repodata.

<details>
<summary>Click for diff
</summary>

```
patching repodata:  11%|█         | 1/9 [02:47<22:21, 167.63s/it]
noarch


patching repodata:  22%|██▏       | 2/9 [02:48<08:05, 69.30s/it] 
linux-armv7l


patching repodata:  33%|███▎      | 3/9 [05:18<10:39, 106.52s/it]
linux-aarch64


linux-64


patching repodata:  44%|████▍     | 4/9 [07:13<09:08, 109.72s/it]
patching repodata:  56%|█████▌    | 5/9 [07:33<05:09, 77.39s/it] 
linux-ppc64le


patching repodata:  67%|██████▋   | 6/9 [09:42<04:44, 94.79s/it]
osx-arm64


patching repodata:  78%|███████▊  | 7/9 [10:02<02:20, 70.49s/it]
win-32


patching repodata:  89%|████████▉ | 8/9 [13:28<01:53, 113.58s/it]
osx-64


win-64


patching repodata: 100%|██████████| 9/9 [14:40<00:00, 100.52s/it]
patching repodata: 100%|██████████| 9/9 [14:40<00:00, 97.81s/it] 
```

</details>

jdblischak:
@conda-forge/core @xhochy very minor PR ready for review

hmaarrfk:
More specificity is better. Why remove specificity if there is no change in existing behavior

jdblischak:
> More specificity is better. Why remove specificity if there is no change in existing behavior

In this case, I find the added specificity to be confusing. The patch needed to be applied to all conda binaries with version less than 0.3.1, regardless of their build number. The patch as it is currently written only worked by mistake (I got lucky since all the existing binaries happened to have build number 0).

In general, I think combining `version_lt` and `build_number` would almost always be a bad idea. I think you should only specify the build number when you are targeting one specific version, not a range of versions.

That being said, I don't feel that strongly about this PR. I'm just trying to be a conscientious contributor and clean up after myself. If you don't want to merge this PR, please go ahead and close it.

hmaarrfk:
The timestamp and less than would stop future builds of old versions in case anybody needs them.

PR #841, opened 2024-09-06
add run_constrains on major compiler version for fenics-basix

minrk:
Checklist

* [x] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [x] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [x] Ran `python show_diff.py` and posted the output as part of the PR.
* [x] Modifications won't affect packages built in the future. 

Packages built with nanobind and pybind11 have a constraint that any other package consuming the binding interface must also use the same major compiler version, which is baked into the ABI.

Only affects fenics-basix 0.8.0 builds 0-2 because 0.7 had a `fenics-basix-pybind11-abi` package to address exactly this. I mistakenly thought nanobind didn't have this problem and removed it for 0.8, but it does and https://github.com/conda-forge/fenics-basix-feedstock/pull/21 puts back an equivalent abi package for the nanobind abi in fenics-basix 0.8.0 starting with build 3.

<details>
<summary>diff</summary>

```
================================================================================
osx-arm64
osx-arm64::fenics-basix-0.8.0-py311h8658ff4_0.conda
osx-arm64::fenics-basix-0.8.0-py311h6bde47b_1.conda
osx-arm64::fenics-basix-0.8.0-py310he1a186f_1.conda
osx-arm64::fenics-basix-0.8.0-py312h157fec4_1.conda
osx-arm64::fenics-basix-0.8.0-py312h69e22ea_0.conda
osx-arm64::fenics-basix-0.8.0-py39ha1e04a5_1.conda
osx-arm64::fenics-basix-0.8.0-py39h821966d_0.conda
osx-arm64::fenics-basix-0.8.0-py310hbfc717c_0.conda
+    "clangxx 16.*",
osx-arm64::fenics-basix-0.8.0-py312h6142ec9_2.conda
osx-arm64::fenics-basix-0.8.0-py310h7306fd8_2.conda
osx-arm64::fenics-basix-0.8.0-py313hf9c7212_2.conda
osx-arm64::fenics-basix-0.8.0-py39h157d57c_2.conda
osx-arm64::fenics-basix-0.8.0-py311h2c37856_2.conda
+    "clangxx 17.*",
================================================================================
================================================================================
linux-ppc64le
linux-ppc64le::fenics-basix-0.8.0-py312hc0892ed_2.conda
[...]

PR #845, opened 2024-09-13
Revert "add run_constrains on major compiler version for fenics-basix"

beckermr:
Reverts conda-forge/conda-forge-repodata-patches-feedstock#841


beckermr:
cc @minrk

minrk (approved):

beckermr:
Well there are two things here. The patch file is wrong, but also there appears to be an issue with changes in pydantic or something.

beckermr:
The interesting bit is that the change wasn't generating valid json which means this is an upstream bug. Hopefully if we skip 2.9 it will get fixed eventually. It could be in json schema or pydantic.

beckermr:
I am going to close this and submit two PRs that fix each thing instead of reverting.

beckermr:
xref: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/846

Note there is no bug in the patch from Min as he correctly pointed out!

PR #865, opened 2024-09-27
chromadb depends on uvicorn-standard not uvicorn

jamesmyatt:
chromadb depends on uvicorn-standard not uvicorn
Fixed in https://github.com/conda-forge/chromadb-feedstock/pull/51

Checklist

### DIFF
--- a/recipe/patch_yaml/chromadb.yaml
+++ b/recipe/patch_yaml/chromadb.yaml
@@ -1,0 +1,9 @@
+# chromadb depends on uvicorn-standard not uvicorn
+# Fixed in https://github.com/conda-forge/chromadb-feedstock/pull/51
+if:
+  name: chromadb
+  timestamp_lt: 1727439944000
+then:
+  - replace_depends:
+      old: uvicorn
+      new: uvicorn-standard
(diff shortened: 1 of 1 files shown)


xhochy:
Why is this still in draft? It's probably good to go.

jamesmyatt:
I haven't done `python show_diff.py` yet. I was having trouble with my environment earlier.

xhochy:
CI runs it, you could theoretically copy from there. Sadly, there the diff is empty.

recipe/patch_yaml/chromadb.yaml
```diff
@@ -0,0 +1,9 @@
+# chromadb depends on uvicorn-standard not uvicorn
+# Fixed in https://github.com/conda-forge/chromadb-feedstock/pull/51
+if:
+  name: chromadb
+  timestamp_lt: 1727439944000
+then:
+  - replace_depends:
+      old: uvicorn
+      new: uvicorn-standard
```
xhochy, line 9: These are exact matches, we need some kind of pattern matching here.

jamesmyatt: Ah. Should be "rename_<>" not "replace_<>".

xhochy (approved):

xhochy:
Thank you!

PR #878, opened 2024-10-13
use a swapfile to help with memory pressure

h-vetinari:
I see a lot of recent jobs are hard on the memory limits of the CI agents and occasionally fail. Let's add a swap file to fix this.

We could also free up some disk space, but 🤞 that we don't need to spend that time; c.f. https://github.com/conda-forge/cdt-builds/commit/2e36c633696ad2f82f50b9abf0d542f22c42daf8

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -25,6 +25,13 @@ jobs:
     displayName: Configure binfmt_misc
 
   - script: |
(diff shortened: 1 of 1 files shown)


hmaarrfk:
should this be an opton.

h-vetinari:
> should this be an opton???

It's an option in smithy already, but we don't use smithy here. On this feedstock, the base workload seems to push hard against the memory limit (and that's unlikely to change, except perhaps once things change to sharded repodata), so I don't think we need to make this an option.

PR #890, opened 2024-10-28
Patch simpeg dependencies

jcapriot:
Patch simpeg for numpy upper bounds, next version will support 2.0

Checklist


hmaarrfk:
Upper bound was added to recipe: https://github.com/conda-forge/simpeg-feedstock/blob/main/recipe/meta.yaml#L24

hmaarrfk:
i feel like azure isn't running.... we have to wait...

jcapriot:
Silly me using the `.yml` extension... It should be good to go now.

PR #913, opened 2024-12-02
Pin cryptography to versions compatible with pyopenssl

IzerOnadimQC:
Checklist

* [x] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [x] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.
* [x] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [x] Ran `python show_diff.py` and posted the output as part of the PR.
* [x] Modifications won't affect packages built in the future. 

<details>
<summary>Diff output</summary>

```
$ python show_diff.py
================================================================================
================================================================================
linux-armv7l
================================================================================
================================================================================
win-32
win-32::pyopenssl-16.2.0-py27_0.tar.bz2
win-32::pyopenssl-16.2.0-py35_0.tar.bz2
win-32::pyopenssl-16.2.0-py36_0.tar.bz2
-    "cryptography >=1.3.4",
+    "cryptography >=1.3.4,<42.0.0a0",
win-32::pyopenssl-17.2.0-py27_0.tar.bz2
win-32::pyopenssl-17.2.0-py35_0.tar.bz2
win-32::pyopenssl-17.2.0-py36_0.tar.bz2
-    "cryptography >=1.9",
+    "cryptography >=1.9,<42.0.0a0",
win-32::pyopenssl-16.0.0-py27_0.tar.bz2
win-32::pyopenssl-16.0.0-py34_0.tar.bz2
win-32::pyopenssl-16.0.0-py35_0.tar.bz2
-    "cryptography >=1.3",
+    "cryptography >=1.3,<42.0.0a0",
win-32::pyopenssl-17.5.0-py27_1.tar.bz2
win-32::pyopenssl-17.5.0-py35_1.tar.bz2
win-32::pyopenssl-17.5.0-py36_1.tar.bz2
-    "cryptography >=2.1.4",
+    "cryptography >=2.1.4,<42.0.0a0",
================================================================================
================================================================================
linux-ppc64le
linux-ppc64le::pyopenssl-19.0.0-py27_0.tar.bz2
linux-ppc64le::pyopenssl-19.0.0-py36_0.tar.bz2
linux-ppc64le::pyopenssl-19.0.0-py37_0.tar.bz2
linux-ppc64le::pyopenssl-19.0.0-py38_0.tar.bz2
-    "cryptography >=2.2.1",
[...]

PR #919, opened 2024-12-03
(Comments only) Add more context about cryptography pins in pyopenssl

maresb:
Implements https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/913#issuecomment-2513876246

I cherry-picked @IzerOnadimQC's comments and then adapted my comment to explain the rationale for maintaining monotonicity. CC @ocefpaf 

I only modify comments, I don't change anything consequential.

### DIFF
--- a/recipe/patch_yaml/pyopenssl.yaml
+++ b/recipe/patch_yaml/pyopenssl.yaml
@@ -18,6 +18,14 @@ then:
       old: cryptography >=35.0
       new: cryptography >=35.0,<39
 ---
+# The X509_V_FLAG_NOTIFY_POLICY constant was removed from cryptography in version 42.0.0.
+# xref: https://github.com/pyca/cryptography/pull/9163
+# This constant was used in pyopenssl versions prior to version 23.2.0.
+# xref: https://github.com/pyca/pyopenssl/pull/1213
+# pyopenssl versions prior to 22.0.0 did not have an upper bound on cryptograpy to exclude newer versions.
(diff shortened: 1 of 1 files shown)


IzerOnadimQC:
Thanks!

ocefpaf:
Thanks @IzerOnadimQC for the detailed explanation of the problem!

ReimarBauer:
That is strange now I get
- cryptography                   44.0.0        py311hafd3f86_0          conda-forge
- pyopenssl                      21.0.0        pyhd8ed1ab_0             conda-forge

and it fails again https://github.com/Open-MSS/MSS/actions/runs/12122317805/job/33854056820

maresb:
@ReimarBauer, it's loading the environment from cache. Do you have permissions to delete the cache files? (If not, there are some dirty tricks to bust the cache from the workflow YAML.)

ReimarBauer:
result is and it works again
- cryptography                   43.0.3        py311h544b224_0          conda-forge
- pyopenssl                      24.2.1        pyhd8ed1ab_2             conda-forge

maresb:
> How can I recognize that I need to delete the cache?

Generally you shouldn't have to worry about this, and there are sane defaults for the cache key. The issue in this particular case is that adjusting the repodata is a fairly extreme (although unfortunately not unheard-of) intervention, and there's currently no cache busting mechanism for it. I think that implementing something would be pretty difficult.

PR #929, opened 2024-12-17
Fix version of pip with wrong python constraints

traversaro:
Fix part of https://github.com/conda-forge/pip-feedstock/issues/132 .

Checklist

### DIFF
--- a/recipe/patch_yaml/pip.yaml
+++ b/recipe/patch_yaml/pip.yaml
@@ -42,3 +42,15 @@ then:
   - add_depends:
       - setuptools
       - wheel
+---
+# See https://github.com/conda-forge/pip-feedstock/issues/132
+# There is a build of pip that does not have wheel and setuptools
+# dependency, and does not have the correct constraint on the python version
+if:
+  name: pip
+  version_in: ["24.3.1"]
+  build: pyh145f28c_1
+then:
+  - replace_depends:
+      old: python >=3.9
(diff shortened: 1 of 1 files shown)


recipe/patch_yaml/pip.yaml
```diff
@@ -42,3 +42,15 @@ then:
   - add_depends:
       - setuptools
       - wheel
+---
+# See https://github.com/conda-forge/pip-feedstock/issues/132
+# There is a build of pip that does not have wheel and setuptools
+# dependency, and does not have the correct constraint on the python version
```
h-vetinari, line 48: Should we not add wheel and setuptools then?

traversaro: It is correct for pip not to have wheel and setuptools if installed with `python>=3.13`, see https://github.com/conda-forge/pip-feedstock/pull/126 . The problem is that this build was meant to be installed only for  `python>=3.13`, but this specific build did not have this constraint. This repodata patch remove the incorrect run constraint `python >=3.9` and adds the correct one `python >=3.13.0a0` (see also the output of `show_diff` script).

h-vetinari: OK, you're right, I had gotten it mixed up how things _should_ have been vs. how they were vs. how to undo the damage. 😅

h-vetinari (approved):

PR #939, opened 2025-01-10
parcels: zarr<3

VeckoTheGecko:
https://github.com/conda-forge/parcels-feedstock/pull/129#issuecomment-2582588956

Checklist

### DIFF
--- a/recipe/patch_yaml/parcels.yaml
+++ b/recipe/patch_yaml/parcels.yaml
@@ -8,3 +8,15 @@ then:
   - replace_depends:
       old: zarr >=2.11.0
       new: zarr >=2.11.0,!=2.18.0
+
+---
+
+# https://github.com/conda-forge/parcels-feedstock/pull/129
+if:
+  name: parcels
+  version: 3.1.1
+  build_number_in: [0, 1]
+then:
+  - replace_depends:
+      old: zarr >=2.11.0,!=2.18.0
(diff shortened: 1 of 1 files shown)


recipe/patch_yaml/parcels.yaml
```diff
@@ -8,3 +8,16 @@ then:
   - replace_depends:
       old: zarr >=2.11.0
       new: zarr >=2.11.0,!=2.18.0
+
+---
+
+# https://github.com/conda-forge/parcels-feedstock/pull/730
```
ocefpaf, line 14: This is a 404.

VeckoTheGecko: Picked the wrong issue number. Fixed now

VeckoTheGecko:
@ocefpaf Is it an issue that this doesn't also target earlier versions of parcels? Or is it a necessary evil (i.e., that you cant retrospectively add upperbounds for everything as the software sphere moves forward).

ocefpaf:
> @ocefpaf Is it an issue that this doesn't also target earlier versions of parcels? Or is it a necessary evil (i.e., that you cant retrospectively add upperbounds for everything as the software sphere moves forward).

This is kind of a choice of how much patching you want. IMO, patching the latest/most used version is essential, anything beyond that may be too much. Sure, those old version may be broken, but the promise of support is probably long gone, people who really need them can figure out how to downgrade zarr, etc.

PR #947, opened 2025-01-16
make latest conda-build repodata patch also restrict lief 0.15

tkelman:
Checklist

* [x] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [ ] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.
* [ ] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [ ] Ran `python show_diff.py` and posted the output as part of the PR.
* [x] Modifications won't affect packages built in the future. 

ref #946 and https://github.com/conda/conda-build/issues/5594

I don't have a local environment set up at the moment to immediately do the `show_diff.py` checklist step

(alternative version is 05f434ad589952d30f74bcfff8d1d741adcd3d33)

PR #948, opened 2025-01-17
expand lief <0.15 patch to catch latest conda-build 25.1.1

minrk:
#947 tightened the patch, but the latest build of conda-build 25.1.1 (build 1), which only pinned <0.16, was published after the timestamp_le cutoff, so mac builds are still running with py-lief 0.15 and failing

### DIFF
--- a/recipe/patch_yaml/conda-build.yaml
+++ b/recipe/patch_yaml/conda-build.yaml
@@ -115,8 +115,8 @@ then:
 # https://github.com/conda/conda-build/issues/5594
 if:
   name: conda-build
-  version_le: 25.11.1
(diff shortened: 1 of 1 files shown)


xhochy (approved):

tkelman:
Thanks for catching this. I also didn't notice that "25.11.1" was a typo.

CI on main failed due to what I think is an http error, could someone restart it? Likewise a lot of the latest CI runs on the conda-build feedstock timed out, that were trying to deploy a new build of 25.11.1 with a `py-lief <0.15` constraint

xhochy:
Triggered a rerun and that succeeded.

PR #955, opened 2025-01-31
fix: patch rubin-sim 2.2.0 for numpy 2

beckermr:
Checklist

xref: https://github.com/conda-forge/rubin-sim-feedstock/pull/59

### DIFF
--- a/recipe/patch_yaml/rubin_sim.yaml
+++ b/recipe/patch_yaml/rubin_sim.yaml
@@ -5,4 +5,4 @@ if:
 then:
   - replace_depends:
       old: numpy
-      new: numpy >= 2
+      new: numpy >=2
--- a/recipe/patch_yaml/rubin_sim.yaml
+++ b/recipe/patch_yaml/rubin_sim.yaml
@@ -1,0 +1,8 @@
+if:
+  name: rubin-sim
+  version: 2.2.0
+  build_number_in: [0]
+then:
+  - replace_depends:
+      old: numpy
+      new: numpy >= 2


recipe/patch_yaml/rubin_sim.yaml
```diff
@@ -0,0 +1,8 @@
+if:
+  name: rubin-sim
+  version: 2.2.0
+  build_number_in: [0]
+then:
+  - replace_depends:
+      old: numpy
+      new: numpy >= 2
```
beckermr, line 8:
```suggestion
      new: numpy >=2
```

jakirkham:
Thanks Matt! 🙏

It looks like PR ( https://github.com/conda-forge/rubin-sim-feedstock/pull/59 ) is still open. Do you plan to merge that?

Recall we have usually made the fix in the latest build and repodata patched older ones

beckermr:
diff is correct

```
noarch::rubin-sim-2.2.0-pyhd8ed1ab_0.conda
-    "numpy",
+    "numpy >=2",
```

beckermr:
I will leave the merging of that PR up to @rhiannonlynne. The next version of this package doesn't need the numpy >=2 constraint and so unless we need another build of 2.2.0, there is no reason to merge it.

jakirkham:
Ok at least my understanding of why we add the pin to the recipe is in the event the same version of the package needs to be rebuilt for another reason. Having the pin there ensures we don't forget this constraint when addressing a different issue. Something that is especially helpful if there is time pressure

beckermr:
Right except that for the next version of the code, this pin will be wrong and we'll need to remove it again. IDK what is best to do here. It is a one-time case I think.

jakirkham:
It would be ok to drop it then

Would Grayskull pick up this change for you?

PR #962, opened 2025-02-19
stricter upper pin on dep wxwidgets

mahf708:
Checklist

xref https://github.com/conda-forge/wxwidgets-feedstock/pull/50
xref https://github.com/conda-forge/ncvis-climate-feedstock/issues/15
xref https://github.com/wxWidgets/wxWidgets/issues/25173

### DIFF
--- a/recipe/patch_yaml/ncvis-climate.yaml
+++ b/recipe/patch_yaml/ncvis-climate.yaml
@@ -1,6 +1,6 @@
 if:
   name: ncvis-climate
-  timestamp_lt: 1739760494852 # Mon Feb 17 02:48:49 2025
+  timestamp_lt: 1739760494852  # Mon Feb 17 02:48:49 2025
 then:
   - tighten_depends:
       name: wxwidgets
--- a/recipe/patch_yaml/ncvis-climate.yaml
+++ b/recipe/patch_yaml/ncvis-climate.yaml
@@ -1,7 +1,0 @@
-if:
-  name: ncvis-climate
-  timestamp_lt: 1739760494852  # Mon Feb 17 02:48:49 2025
-then:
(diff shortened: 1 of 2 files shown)


recipe/patch_yaml/ncvis-climate.yaml
```diff
@@ -0,0 +1,7 @@
+if:
+  name: ncvis-climate
+  timestamp_lt: 1739760494852  # Mon Feb 17 02:48:49 2025
+then:
+  - tighten_depends:
+      name: wxwidgets
+      max_pin: 'x.x.x'
```
hmaarrfk, line 7: please lets only do this ">=3.2.5"

hmaarrfk: see https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/patch_yaml/numba.yaml for examples

recipe/patch_yaml/wxwidgets.yaml
```diff
@@ -0,0 +1,9 @@
+# wxwidgets broke abi compatiblity in 3.2.6
+# see issue https://github.com/wxWidgets/wxWidgets/issues/25173
+# so we are tightening the upper bin going forward, starting with 3.2.5
+if:
+  has_depends: wxwidgets >=3.2.5,<3.3.0a0
```
hmaarrfk, line 5: You also need the timestamp limit too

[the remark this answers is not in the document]
hmaarrfk: Yeah we can "extend" it later when some1 verifies this. Eventually this will come to pass.

PR #976, opened 2025-03-12
patch builds built with fmt 11.1 to use 11.0

h-vetinari:
Checklist

* [x] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [x] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.
* [x] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [x] Ran `python show_diff.py` and posted the output as part of the PR.
* [x] Modifications won't affect packages built in the future. 

For https://github.com/conda-forge/admin-requests/pull/1433 / https://github.com/conda-forge/fmt-feedstock/issues/61

## Output of `python show_diff.py`

```diff
>python show_diff.py
================================================================================
================================================================================
linux-armv7l
================================================================================
================================================================================
win-32
================================================================================
================================================================================
linux-ppc64le::dolfinx_mpc-0.9.0-py310h3e2241c_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py310h7a7733f_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py310h9e9f3c9_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py310ha2969e0_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py311h2303d79_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py311h35e8972_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py311h9ad1846_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py311hf7b3e08_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py312h3e7125a_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py312h782b267_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py312h810358a_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py312he5b6e0a_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py313h12d4a3b_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py313h973691d_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py313h9f8e710_3.conda
linux-ppc64le::dolfinx_mpc-0.9.0-py313hfee17fd_3.conda
[...]

PR #979, opened 2025-03-14
Revert "patch builds built with fmt 11.1 to use 11.0"

minrk:
Reverts conda-forge/conda-forge-repodata-patches-feedstock#976

this patch led to [broken envs](https://github.com/conda-forge/dolfinx_mpc-feedstock/issues/27) because (at least some) builds with 11.1 depend on symbols not in 11.0. It seems better to me to leave these builds unsolvable than try to put them into broken envs.

### DIFF
--- a/recipe/patch_yaml/fmt.yaml
+++ b/recipe/patch_yaml/fmt.yaml
@@ -1,9 +1,0 @@
-# see https://github.com/conda-forge/fmt-feedstock/issues/61
-# and https://github.com/conda-forge/admin-requests/pull/1433
-if:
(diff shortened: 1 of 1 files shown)


minrk:
Marking all of the affected packages broken is okay, too, but I'm not sure there's a reason to unless future 11.1 releases will be ABI-incompatible with these.

h-vetinari:
I would have just marked the dolfinx_mpc builds as broken, because this is the outcome now in any case: all builds depending on >=11.1 need to be marked broken.

PR #995, opened 2025-04-05
Relax libgfortran patch to change libgfortran version scheme on osx to match linux

isuruf:
Checklist

To allow https://github.com/conda-forge/gfortran_impl_osx-64-feedstock/pull/89

### DIFF
--- a/recipe/patch_yaml/libgfortran.yaml
+++ b/recipe/patch_yaml/libgfortran.yaml
@@ -1,0 +1,7 @@
+if:
+  timestamp_lt: 1743886623000
+  has_depends: libgfortran 5.*
+then:
+  - replace_depends:
+      old: libgfortran 5.*
+      new: libgfortran >=5


recipe/patch_yaml/libgfortran.yaml
```diff
@@ -0,0 +1,7 @@
+if:
+  timestamp_lt: 1743886623000
+  has_depends: libgfortran 5.*
+then:
+  - replace_depends:
+      old: libgfortran 5.*
+      new: libgfortran >=5
```
beckermr, line 7: Why not replace this with `libgfortran5 >=<some correct lower bound>`?

beckermr (changes requested):

recipe/patch_yaml/libgfortran.yaml
isuruf: We want both `libgfortran` to ensure mutual exclusivity.

beckermr: Ok. I don't quite follow any more since it's been too long since I've worked on this bit of the infrastructure.

beckermr (approved):

PR #993, opened 2025-04-04
Limit libxml2 to <2.14

isuruf:
Checklist

* [ ] Used a static YAML file for the patch if possible ([instructions](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/main/recipe/README.md)).
* [ ] Only wrote code directly into `generate_patch_json.py` if absolutely necessary.
* [ ] Ran `pre-commit run -a` and ensured all files pass the linting checks.
* [ ] Ran `python show_diff.py` and posted the output as part of the PR.
* [ ] Modifications won't affect packages built in the future. 

See https://github.com/conda-forge/libxml2-feedstock/issues/145

PR #998, opened 2025-04-07
adjust libxml bounds to only affect <3.0 bound

msarahan:
#996 was still not a complete fix. I don't understand why. Nevertheless, the simplified code here leaves more of the original pins alone. There were many pins on minor versions that we were overwriting with 2.14, and I'm not sure that they were ok.

There's a huge number of changes reported by show_diff.py. They fall into two categories:

This one is fixing the 2.14 bound, which is likely the most pressing issue
```
linux-64::clang-tools-20.1.2-default_hb5137d0_0.conda
-    "libxml2 >=2.13.7,<2.14.0a0"
+    "libxml2 >=2.13.7,<3.0a0"
```

These are reverting the change from #993, preserving the existing x.x pin
```
linux-64::mdal-1.0.1-hba616fa_7.conda
-    "libxml2 >=2.10.4,<2.14.0a0"
+    "libxml2 >=2.10.4,<2.11.0a0"
linux-64::mdal-1.1.0-hccb4986_1.conda
-    "libxml2 >=2.12.2,<2.14.0a0"
+    "libxml2 >=2.12.2,<2.13.0a0"
```

### DIFF
--- a/recipe/patch_yaml/libxml2.yaml
+++ b/recipe/patch_yaml/libxml2.yaml
@@ -11,7 +11,7 @@ if:
 then:
   - tighten_depends:
       name: libxml2
-      upper_bound: 2.11.0
+      upper_bound: "2.11.0"
 ---
 if:
   timestamp_lt: 1744044827000
@@ -20,4 +20,4 @@ if:
 then:
   - tighten_depends:
       name: libxml2
-      upper_bound: 2.14
+      upper_bound: "2.14"
--- a/recipe/patch_yaml/libxml2.yaml
+++ b/recipe/patch_yaml/libxml2.yaml
@@ -7,42 +7,15 @@ if:
   has_depends:
     - libxml2 >=2.9*
 then:
-  - loosen_depends:
+  - tighten_depends:
       name: libxml2
-      upper_bound: 2.14.0
+      upper_bound: 2.11.0
 ---
 if:
   timestamp_lt: 1744044827000
   has_depends:
-    - libxml2 >=2.10*
-then:
-  - loosen_depends:
-      name: libxml2
-      upper_bound: 2.14.0
----
-if:
-  timestamp_lt: 1744044827000
-  has_depends:
-    - libxml2 >=2.11*
-then:
-  - loosen_depends:
-      name: libxml2
-      upper_bound: 2.14.0
----
-if:
-  timestamp_lt: 1744044827000
-  has_depends:
-    - libxml2 >=2.12*
-then:
-  - loosen_depends:
-      name: libxml2
-      upper_bound: 2.14.0
----
-if:
-  timestamp_lt: 1744044827000
-  has_depends:
-    - libxml2 >=2.13*
+    - libxml2 <3.0a0
 then:
   - tighten_depends:
       name: libxml2
-      upper_bound: 2.14.0
+      upper_bound: 2.14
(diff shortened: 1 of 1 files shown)


msarahan:
@conda-forge/core I think this is ready. I'm tempted to merge it since I think it will fix lots of broken projects, but it's also very late in the day for any unintended consequences.

h-vetinari:
> These are reverting the change from #993, preserving the existing x.x pin

That's not correct, they were loosened because libxml 2.11/2.12 etc. were compatible with 2.13, which is still the case. At the time the minor pin was lifted (so they became <3.0a0), and #993 correctly tightened this so <2.14a0 after 2.14 broke the ABI. But that doesn't mean we should retighten the pins before 2.14. That will lead to lots of environments that have been created since then becoming unresolvable.

Further, this still contains changes that incorrectly _loosen_ pins, e.g.
```diff
linux-aarch64::xrootd-5.8.0-py39hf37f3b5_0.conda
-    "libxml2 <2.14.0a0",
-    "libxml2 >=2.13.6,<2.14.0a0",
+    "libxml2",
+    "libxml2 >=2.13.6,<3.0a0",
```

Finally, it patches way too many packages at once (~38'000). I don't know the hard limits where stuff might start going wrong, but when we did the zlib 1.3 patches, @beckermr asked that we keep each batch to maximum ~5000 artefacts.

msarahan:
Thanks for pointing this out. I hadn't looked carefully enough at #993. I will preserve that functionality.

As for changing too many packages at once, I'm not sure what the reasoning is. On one hand, you don't want to impact too many things at one time, just to limit the blast radius if things go wrong. But the CDN argument (what I could gather from the linked thread) doesn't make sense to me. The patches are just one JSON file being merged into another. The idea of batches does not make sense, because no changes are applied to individual packages. It is only a mess of changes in one shot to repodata.json. This said, I'm fine to do it if it is what needs to happen to feel like we're not taking too much risk at once.

msarahan:
I found ABI breakage in a few different places. I'm not deeply involved with any project that uses libxml2, so you probably know better than me. Still, this gives me pause on relaxing those constraints quite so much:

* [2.10](https://github.com/coreos/fedora-coreos-tracker/issues/1336)
* [2.11](https://gitlab.gnome.org/GNOME/libxml2/-/issues/526)
* [2.12](https://bugs-devel.debian.org/cgi-bin/bugreport.cgi?bug=1073508)
* [2.13](https://bugs.gentoo.org/935452)

In each of these, it seems like the libxml2 author was somehow able to work around the incompatibilities. We might be at a steady state of sorts with each of these versions having matured past their compatibility. I definitely understand the cost of pinning things tightly - either unsolvable environments or tons of rebuilds in migrations. I trust your experience and sense for this, but I personally would not unpin any of these beyond `x.x` without extensive integration testing.

h-vetinari:
> I trust your experience and sense for this, but I personally would not unpin any of these beyond `x.x` without extensive integration testing.

TBH I was not involved with the loosening of the pins for libxml2, so I cannot really comment there. The only thing I want to avoid right now is to mix the fixing of our current problems with overhauling past decisions, because IMO those two issues operate on separate time scales. I want to stop the bleeding for 2.14 first and foremost, and then we can talk about whether previous 2.x pins should have been loosened at all (and what - if anything - we should do about it going forward).

FWIW I tend to agree with you that this is risky business (and for zlib we did a pretty in-depth [analysis](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5443#issuecomment-2127447042) from first principles to justify that).

h-vetinari:
PS: And thanks for the explanation how the repodata patching actually works! 🙏 

@beckermr will have to explain what exactly he was/is worried about could happen, I was just following his guidance there.

PR #999, opened 2025-04-08
Tighten `libxml2<3.0a0` bounds to 2.14 where appropriate

h-vetinari:
This patch builds on top of #998.

PR #1001, opened 2025-04-08
libxml 2.12 needs both `{loosen,tighten}_depends`

### LINKED ISSUE #1000: `has_depends` pattern not picking up all relevant builds?
One thing I don't understand is why these artefacts with (e.g.) `libxml2 >=2.12.5,<3.0a0` don't seem to fall under
https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/eeda1291413a5d394cee4310c575ee4ba35d64c0/recipe/patch_yaml/libxml2.yaml#L34-L35

but at least they are being picked up by the reverse condition (`has_depends: [libxml2 *<3.0a0]`).

_Originally posted by @h-vetinari in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/issues/999#issuecomment-2785037432_

h-vetinari:
Follow-up to #999 

Closes #1000 

This is because https://github.com/conda-forge/libxml2-feedstock/pull/111 landed during the libxml2 2.12.x cycle, so builds before that PR need `loosen_depends` (in order to allow libxml 2.13, which is compatible), whereas builds after that PR need `tighten_depends` to avoid https://github.com/conda-forge/libxml2-feedstock/issues/145.

### DIFF
--- a/recipe/patch_yaml/libxml2.yaml
+++ b/recipe/patch_yaml/libxml2.yaml
@@ -36,7 +36,18 @@ if:
   has_depends:
     - libxml2 >=2.12*
 then:
+  # loosen for builds against libxml 2.12 before https://github.com/conda-forge/libxml2-feedstock/pull/111
   - loosen_depends:
+      name: libxml2
+      upper_bound: 2.14.0
+---
+if:
+  timestamp_lt: 1744044827000
+  has_depends:
+    - libxml2 >=2.12*
+then:
+  # tighten for builds against libxml 2.12 after https://github.com/conda-forge/libxml2-feedstock/pull/111
+  - tighten_depends:
       name: libxml2
       upper_bound: 2.14.0
 ---


h-vetinari:
With this PR we can then actually completely remove
https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/2340c9ca03d8fd15d24f7fdcd7c4804269af2126/recipe/patch_yaml/libxml2.yaml#L52-L63
(I tested that it generates the same diff), the only reason I haven't done so yet is because then git creates a very confusing diff for this PR.

h-vetinari:
OK, I ran `show_diff.py` with the changes from https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/1001/commits/10bad89b667691c9b227647f7701899442ba1ce7, and it's empty as it should be.
```
linux-armv7l
win-32
linux-ppc64le
osx-arm64
linux-aarch64
noarch
win-64
osx-64
linux-64
```

Since this doesn't change any package constraints, I'll merge it as a clean-up of the patch_yaml files.

msarahan:
We're all fixed on our side. Thanks for your work on this, @h-vetinari

msarahan:
xref: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/issues/1003

PR #1004, opened 2025-04-08
implement change_depends patch mode for simpler absolute pins

### LINKED ISSUE #1003: YAML syntax for changing bounds / confusion with tighten/loosen
### Comment:

In #998, I was clearly confused regarding the workings of loosen_depends and tighten_depends. The functionality feels especially conflated when used with upper_bound. It seems like I'm not the only one, with @isuruf mixing things up in #994.

I am NOT proposing removing loosen_depends and tighten_depends. I think there's too many existing YAML specs that already use them. I would like to propose a new key at the level of loosen_depends and tighten_depends, though: `change_depends`

Schema:
```
then:
  change_depends:
    lower_bound: <str>, evaluates as >=str
    upper_bound: <str>, evaluates as <str
    exclusions: <list[str]>, evaluates as comma-separated != constraints, with wildcard suffixes where not provided
```

I would also deprecate the use of upper_bound and lower_bound with loosen_depends and tighten_depends, and codify loosen_depends and tighten_depends only as relative operations on the current bounds.

@beckermr, @jaimergp - it seems like you two are the main instigators on the YAML stuff (which I very much like, by the way). Thoughts? If this seems helpful, I can take on implementing it.

msarahan:
The general idea here is that `loosen_depends` and `tighten_depends` are confusing when combined with absolute version numbers (#1003). When dealing with absolute version numbers, I thought it would be easier to just change constraints more directly.

Closes #1003


recipe/README.md
```diff
@@ -136,25 +136,56 @@ then:
...
       name: matplotlib  # this field can use the fnmatch glob syntax
       # you must give one of max_pin or upper_bound
       # optional way to specify the new maximum pin as 'x', 'x.x', etc.
       max_pin: 'x.x'
-      # optional way to specify upper bound explicitly
+      # [DEPRECATED] optional way to specify upper bound explicitly
       # do not use with `max_pin`
+      #
+      # DEPRECATION NOTE: use `max_pin` for a relative scheme, or use `change_depends` 
+      # instead of `tighten_depends` for an absolute scheme
```
isuruf, line 149: I don't think this should be deprecated. If there's `>=2.0,<2.6` and we want to add an upper bound of `2.10` matching `>=2.*`, this is the only way to keep it `>=2.0,<2.6`, right?

recipe/README.md
```diff
@@ -136,25 +136,56 @@ then:
...
       # optional pinning expression 'x', 'x.x', etc. to set how much looser to make the pin
       max_pin: 'x.x'
-      # optional way to specify upper bound explicitly
+      # [DEPRECATED] optional way to specify upper bound explicitly
       # do not use with `max_pin`
+      #
+      # DEPRECATION NOTE: use `max_pin` for a relative scheme, or use `change_depends` 
+      # instead of `loosen_depends` for an absolute scheme
       upper_bound: 2.0.1
+  - change_depends:
```
beckermr, line 166: This name is too generic. Every operation in this repo changes something. Let's call this "bound_depends" or similar.

beckermr (changes requested):

recipe/README.md
msarahan: I don't think I explained this clearly. If you just want to move the decimal place around the lower bound, use tighten or loosen. 

If you want to change bounds to some arbitrary value, you don't care whether you are moving the decimal place around. #1001 is needlessly complicated because of this notion. If you aren't doing a relative computation on the lower bound, then you don't need to know whether it's tightening or loosening, and indeed, those notions as implemented today leave room for error.

With this PR, if you have `>=2.0,2.6`, you have 3 behaviors that I think are clear:

```
tighten_depends:
  max_pin: 'x.x.x' (could make this optional to default to just one more place than current pin)

>=2.0,<2.6.0a0
```

```
loosen_depends:
  max_pin: 'x' (could make this optional to default to just one less place than current pin)

>=2.0,<3.0a0
```

change upper bound: 
```
less_generic_name_depends:
  upper_bound: 2.10

>=2.0,<2.10a0
```

The point is that if you are trying to communicate both relative (decimal shifting) and absolute versions, it's bound to get confusing. I think they should be conceptually separate.

recipe/patch_yaml_model.py
```diff
@@ -196,9 +212,14 @@ class _ThenClauseItem(_ForbidExtra):
...
         description="Make a dependency version constraint looser",
     )
+    change_depends: _Name_LowerBound_UpperBound_Other = Field(
+        None,
+        description="Change a dependency version constraint",
+    )
 
 
 class PatchYaml(_ForbidExtra):
+    model_config = ConfigDict(coerce_numbers_to_str=True)
```
jaimergp, line 222: I'm not sure if this `model_config` assignment overrides the one in the `_ForbidExtra` baseclass.

msarahan: I think you're probably right. This change is not related to the focus of this PR, anyway. I just hit a rather opaque error about schema problems, and the issue was a version like `1.10` being interpreted as a float and triggering the schema violation. Coercing to strings seems like a good idea, but also a subject for another PR.

PR #1008, opened 2025-04-09
Patch packages built against libxml2 2.13

SylvainCorlay:
This is a follow up on #999.

Checklist

### DIFF
--- a/recipe/patch_yaml/libxml2.yaml
+++ b/recipe/patch_yaml/libxml2.yaml
@@ -44,18 +44,9 @@ then:
 if:
   timestamp_lt: 1744044827000
   has_depends:
-    - libxml2 >=2.12*
-then:
-  # tighten for builds against libxml 2.12 after https://github.com/conda-forge/libxml2-feedstock/pull/111
-  - tighten_depends:
-      name: libxml2
-      upper_bound: 2.14.0
----
-if:
-  timestamp_lt: 1744044827000
-  has_depends:
     - libxml2 >=2.13*
 then:
-  - tighten_depends:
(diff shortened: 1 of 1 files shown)


h-vetinari (changes requested):
This makes no sense to me. https://github.com/conda-forge/libxml2-feedstock/pull/111 landed for 2.12, so we definitely need both tighen and loosen for 2.12. Before we touch this any further, we should look at the exact constraints of builds that still have been missed, and why they don't fall under any of the existing instructions here.

h-vetinari:
See [here](https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/issues/1000#issuecomment-2791058176). The pull request template already asks for it, but in this case we really really need to see the output of `python show_diff.py` before we make any further changes that end up (re)breaking yet another case.

SylvainCorlay:
> This makes no sense to me. https://github.com/conda-forge/libxml2-feedstock/pull/111 landed for 2.12, so we definitely need both tighen and loosen for 2.12. Before we touch this any further, we should look at the exact constraints of builds that still have been missed, and why they don't fall under any of the existing instructions here.

No need to review this at this point (I made the PR draft). I was confused by the `tighten_depends` and `loosen_depends` logic and was planing on looking more into it tomorrow if necessary.

> See https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/issues/1000#issuecomment-2791058176. The pull request template already asks for it, but in this case we really really need to see the output of python show_diff.py before we make any further changes that end up (re)breaking yet another case.

PR #1012, opened 2025-04-13
Loosen ABI compatibility for packages built with qt6.8

hmaarrfk:
see 
* https://code.qt.io/cgit/qt/qtreleasenotes.git/about/qt/6.9.0/release-note.md
* https://code.qt.io/cgit/qt/qtreleasenotes.git/about/qt/6.8.0/release-note.md


hmaarrfk:
@conda-forge/conda-forge-repodata-patches does anybody know why my patching isn't "loosening" the pin to 6.10? 

cc: @jschueller

PR #1016, opened 2025-04-25
Fix h11 dependency in httpcore=0.13.3

xhochy:
Checklist

This was fixed in https://github.com/conda-forge/httpcore-feedstock/pull/30 but needs a patch for the `*_0` builds.

cc @conda-forge/httpcore


xhochy:
Diff:

```
noarch
noarch::httpcore-0.13.3-pyhd8ed1ab_0.tar.bz2
-    "h11 0.*",
+    "h11 >=0.11,<0.13",
```

PR #1021, opened 2025-05-06
Extend xz_split scope

xhochy:
Checklist


xhochy:
I have also updated the timestamp, as this had not been reflected in the feedstock. Now it is: https://github.com/conda-forge/xz-feedstock/pull/53

ocefpaf (approved):

PR #1035, opened 2025-06-05
add geopandas constraint to polartoolkit

mdtanker:
Checklist


mdtanker:
Any ideas why this is failing? I've tried to follow the same syntax of other .yaml files here.

xhochy (approved):

xhochy:
These checks are only run in CI not in pre-commit as they are probably quite expensive.

PR #1037, opened 2025-06-06
Add pyqt constraint against pyqt_builder

JarrettSJohnson:
Checklist

Currently adding a patch to `pyqt-builder` to be used for all `pyqt` v6 versions. All current and new `v5` versions should use `pyqt-builder <1.16.0`. I tried running `show_diff.py` but didn't see anything; so I'll wait for the Azure output.

As advised in https://github.com/conda-forge/pyqt-builder-feedstock/pull/22

@hmaarrfk 

<details>

<summary> Patching output </summary>

```
linux-armv7l

patching repodata:  11%|█         | 1/9 [00:00<00:06,  1.32it/s]

linux-ppc64le

patching repodata:  22%|██▏       | 2/9 [03:16<13:28, 115.47s/it]

linux-aarch64

patching repodata:  33%|███▎      | 3/9 [03:48<07:43, 77.22s/it] 

noarch

patching repodata:  44%|████▍     | 4/9 [04:09<04:36, 55.32s/it]

win-32


patching repodata:  56%|█████▌    | 5/9 [04:33<02:54, 43.70s/it]
osx-arm64


linux-64

patching repodata:  67%|██████▋   | 6/9 [07:22<04:19, 86.39s/it]
patching repodata:  78%|███████▊  | 7/9 [09:08<03:05, 92.70s/it]

win-64

patching repodata:  89%|████████▉ | 8/9 [10:09<01:22, 82.90s/it]

osx-64

patching repodata: 100%|██████████| 9/9 [10:47<00:00, 68.77s/it]
patching repodata: 100%|██████████| 9/9 [10:47<00:00, 71.96s/it]
```

</details>

### DIFF
--- a/recipe/patch_yaml/pyqt.yaml
+++ b/recipe/patch_yaml/pyqt.yaml
@@ -1,0 +1,8 @@
+if:
+  name: pyqt
+  version_lt: 6.0.0
+  timestamp_lt: 1749235324000
+then:
+  - replace_depends:
+      old: pyqt-builder
+      new: pyqt-builder <1.16.0


hmaarrfk:
Do you need to add this constraint to a few similar packages:
https://github.com/search?q=org%3Aconda-forge%20pyqt-builder&type=code

JarrettSJohnson:
Probably. Does that sound right, @ccordoba12 ? Would I also need to submit that pinning to each of these repos, too?

ccordoba12:
> Would I also need to submit that pinning to each of these repos, too?

Yeah, it seems it'd be necessary to do that too (I didn't know more packages relied on `pyqt-builder`).

Given that, I think you should use the pin suggested by @hmaarrfk in your PyQt PR: `pyqt-builder =1.15`.

@hmaarrfk, then I guess the repodata changes here should include those pins too?

JarrettSJohnson:
I'll submit the pin to `1.15` PR to :
- pyqt ( https://github.com/conda-forge/pyqt-feedstock/pull/158 )
- pyqtwebkit
- qscintilla2 
- qgis
- pyqtads

This repodata patch commit, I assume, will have to wait until all of these PRs are merged?

JarrettSJohnson:
Actually, these other feedstocks seem to include both `pyqt` and `pyqt-builder`. Wouldn't the constraints from `pyqt` be enough?

ccordoba12:
> Actually, these other feedstocks seem to include both pyqt and pyqt-builder. Wouldn't the constraints from pyqt be enough?

Yeah, you're right: pyqt-builder is a host dependency of pyqt, not a build one.

hmaarrfk:
> @hmaarrfk, then I guess the repodata changes here should include those pins too?

the goal of the repo data patch is to be the least invasive possible while addressing the incompatibility. 

So here the « less than » is appropriate

hmaarrfk:
sorry i misunderstood. this isn't needed since there is no runtime dependency on pyqt-builder.

only a build time dependency.

this just means you have to "update" your recipes as you did. 

repo-data-patches are only for runtime dependencies.

PR #1067, opened 2025-08-25
Align scipy 1.8.1 dependency constraints with maintenance branch

julian-belina:
Checklist

### DIFF
--- a/recipe/patch_yaml/scipy.yaml
+++ b/recipe/patch_yaml/scipy.yaml
@@ -62,12 +62,12 @@ then:
 if:
   name: scipy
   version: 1.8.1
-  build_number_in: [0, 2]
+  build_number_in: [0, 1, 2]
   timestamp_le: 1686323537000
 then:
   - replace_depends:
       old: numpy >=1.19.5,<2.0a0
-      new: scipy >=1.17.3,<1.25.0
+      new: numpy >=1.19.5,<1.25.0
   - replace_depends:
       old: numpy >=1.21.6,<2.0a0
-      new: scipy >=1.17.3,<1.25.0
+      new: numpy >=1.19.5,<1.25.0
--- a/recipe/patch_yaml/scipy.yaml
+++ b/recipe/patch_yaml/scipy.yaml
@@ -70,4 +70,4 @@ then:
       new: numpy >=1.19.5,<1.25.0
   - replace_depends:
       old: numpy >=1.21.6,<2.0a0
-      new: numpy >=1.19.5,<1.25.0
+      new: numpy >=1.21.6,<1.25.0
--- a/recipe/patch_yaml/scipy.yaml
+++ b/recipe/patch_yaml/scipy.yaml
@@ -53,3 +53,21 @@ if:
 
 then:
   - add_constrains: libopenblas <0.3.26
+
+---
+# Allign constrains with limits in 1.8.x maintenance branch
+# The numpy min and max versions are documented here:
+# https://github.com/scipy/scipy/blob/maintenance/1.8.x/setup.py
+# https://github.com/scipy/scipy/blob/maintenance/1.8.x/scipy/__init__.py
+if:
+  name: scipy
+  version: 1.8.1
+  build_number_in: [0, 2]
+  timestamp_le: 1686323537000
+then:
+  - replace_depends:
+      old: numpy >=1.19.5,<2.0a0
+      new: scipy >=1.17.3,<1.25.0
+  - replace_depends:
+      old: numpy >=1.21.6,<2.0a0
+      new: scipy >=1.17.3,<1.25.0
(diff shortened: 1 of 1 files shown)


julian-belina:
The dependency constraints of the scipy=1.8.1 build number=[0, 2] were different from the required numpy version in the maintenance branch. The correct constraints are >=1.17.3,<1.25.0, which can be found here: 

https://github.com/scipy/scipy/blob/maintenance/1.8.x/setup.py
https://github.com/scipy/scipy/blob/maintenance/1.8.x/scipy/__init__.py

h-vetinari:
numpy 1.8 is [older](https://github.com/numpy/numpy/releases/tag/v1.8.1) than [conda-forge](https://conda-forge.org/blog/2025/04/11/ten-years-of-conda-forge/) - what is anyone still doing with such ancient versions?

recipe/patch_yaml/scipy.yaml
```diff
@@ -53,3 +53,21 @@ if:
...
+
+---
+# Align constrains with limits in 1.8.x maintenance branch
+# The numpy min and max versions are documented here:
+# https://github.com/scipy/scipy/blob/maintenance/1.8.x/setup.py
+# https://github.com/scipy/scipy/blob/maintenance/1.8.x/scipy/__init__.py
+if:
+  name: scipy
+  version: 1.8.1
+  build_number_in: [0, 2]
```
h-vetinari, line 65: this is not a range, you have to provide all
```suggestion
  build_number_in: [0, 1, 2]
```

recipe/patch_yaml/scipy.yaml
```diff
@@ -53,3 +53,21 @@ if:
...
+  timestamp_le: 1686323537000
+then:
+  - replace_depends:
+      old: numpy >=1.19.5,<2.0a0
+      new: scipy >=1.17.3,<1.25.0
```
h-vetinari, line 70: This makes no sense. You're replacing numpy with scipy. And you cannot replace the lower bound, because we were building against 1.19, which is what determines the lower bound.
```suggestion
  - replace_depends:
      old: numpy >=1.19.5,<2.0a0
      new: nympy >=1.19.5,<1.25.0
```

recipe/patch_yaml/scipy.yaml
```diff
@@ -53,3 +53,21 @@ if:
...
+  - replace_depends:
+      old: numpy >=1.21.6,<2.0a0
+      new: scipy >=1.17.3,<1.25.0
```
h-vetinari, line 73: same here

h-vetinari (changes requested):

recipe/patch_yaml/scipy.yaml
```diff
@@ -53,3 +53,21 @@ if:
...
+  version: 1.8.1
+  build_number_in: [0, 1, 2]
+  timestamp_le: 1686323537000
+then:
+  - replace_depends:
+      old: numpy >=1.19.5,<2.0a0
+      new: numpy >=1.19.5,<1.25.0
+  - replace_depends:
+      old: numpy >=1.21.6,<2.0a0
+      new: numpy >=1.19.5,<1.25.0
```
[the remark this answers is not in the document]
h-vetinari: no, it's rather that python 3.10 was only supported from numpy 1.21, so it wasn't possible to use the default numpy version we had at the time (1.19) for the at-the-time newest python

recipe/patch_yaml/scipy.yaml
```diff
@@ -53,3 +53,21 @@ if:
...
+
+---
+# Align constrains with limits in 1.8.x maintenance branch
+# The numpy min and max versions are documented here:
+# https://github.com/scipy/scipy/blob/maintenance/1.8.x/setup.py
+# https://github.com/scipy/scipy/blob/maintenance/1.8.x/scipy/__init__.py
+if:
+  name: scipy
+  version: 1.8.1
+  build_number_in: [0, 1, 2]
```
h-vetinari, line 65: I was surprised that build 1 doesn't show up in the diff, but that's because we apparently went from build number 0 to 2 without ever publishing builds for 1.

h-vetinari (commented):
The utility of this is kind of questionable to me. Yes, the metadata of scipy 1.8 says that it's not compatible with the unknown-at-the-time-because-in-the-future numpy 1.26, but in reality, numpy 1.26 is completely compatible with 1.25. The [release notes](https://github.com/numpy/numpy/releases/tag/v1.26.0) even state
> The NumPy 1.26.0 release is a continuation of the 1.25.x release cycle
with the addition of Python 3.12.0 support.

I did https://github.com/conda-forge/scipy-feedstock/pull/218 because the [issue](https://github.com/conda-forge/scipy-feedstock/issues/216) had been more pronounced for scipy 1.7, and I just applied the fix to all relevant versions at the time.

I'm a bit on the fence whether this case (and patching such old builds in general) is really worth it. Can you explain why you would need this @julian-belina?

julian-belina:
@h-vetinari Thank you for the explanation. I will assume that SciPy 1.8.1 is compatible with NumPy 1.25 and 1.26, and close this pull request.

PR #1097, opened 2025-10-18
Patch numpy dependency for xarray=2023.10.0

julian-belina:
Checklist

### DIFF
--- a/recipe/patch_yaml/xarray.yaml
+++ b/recipe/patch_yaml/xarray.yaml
@@ -94,10 +94,20 @@ then:
       old: pandas >=1.4
       new: pandas >=1.4,<2a0
 ---
-
 # 2024/01 - hmaarrfk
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Can you keep the old replacement too? Or perhaps write it in a way that it would apply to both ordering of patches?

im scared the patch order will flip in the future.

julian-belina:
I'm afraid I'm not familiar enough with the patching process to know if it makes a difference. However, if you think it's useful, I can re-add the old patch.

Do you know where the patch that applied the upper limit came from?

julian-belina:
```
linux-armv7l
win-32
linux-ppc64le
osx-arm64
linux-aarch64
win-64
osx-64
noarch
noarch::xarray-2023.10.0-pyhd8ed1ab_0.conda
-    "numpy >=1.21,<2.0a0",
+    "numpy >=1.22,<2.0a0",
linux-64

```

PR #1098, opened 2025-10-20
Introduce upper limits geopandas >=0.12.1,<=0.14.3 for fiona

julian-belina:
Checklist

### DIFF
--- a/recipe/patch_yaml/geopandas.yaml
+++ b/recipe/patch_yaml/geopandas.yaml
@@ -5,7 +5,7 @@ if:
   name: geopandas
   timestamp_lt: 1760977826000
   version_le: 0.14.3
-  version_gt: 0.1.1
+  version_ge: 0.12.1
 then:
   - tighten_depends:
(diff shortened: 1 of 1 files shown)


recipe/patch_yaml/geopandas.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: geopandas
+  timestamp_lt: 1760977826000
+  version_le: 0.12.2
+  version_ge: 0.12.1
+then:
+  - replace_depends:
+      old: fiona 
+      new: fiona <=1.9.6
+
+
```
h-vetinari: I'm wondering if we should add the `fiona <1.10` cap to all old geopandas release that don't have it. patching 0.12.1 but not 0.12.0 seems suspicious to me. 

The following should work uniformly, without having to match the patch-versions in the fiona lower bound.
```suggestion
then:
  - tighten_depends:
      name: fiona
      upper_bound: 1.10
```

PR #1099, opened 2025-10-21
Fix typo in Readme.md

julian-belina:
Checklist


julian-belina:
I just wanted to point out the typo that was missed in yesterday's PR, @h-vetinari 

There seems to be a self-difference that I can't explain. However, no recipe was altered.

h-vetinari (approved):
Thank you!

PR #1103, opened 2025-10-29
openfe: add run_constrained for dask

mikemhenry:
Checklist


mikemhenry:
With the current CDN issues, things got really wierd when I ran without the `--subdirs noarch` command I think I just got lucky when I ran it with `--subdirs noarch` locally once, I will check the output of the CI job. 
```
$ python show_diff.py --subdirs noarch
noarch
noarch::openfe-1.7.0-pyhd8ed1ab_0.conda
noarch::openfe-1.7.0-pyhd8ed1ab_1.conda
+  "constrains": [
+    "dask >=2025"
+  ],
```

mikemhenry:
Also just to check, this patch is doing the same thing that I did in this PR, right? https://github.com/conda-forge/openfe-feedstock/pull/107/files

hmaarrfk:
it should be, so the question for me is, why is it double applying the patch?

mikemhenry:
I bet it is because of this massive footgun that hits me all the time, `openfe-base` is a separate package from `openfe` and the PR I made put the constraint in for `openfe-base` and not `openfe` (this is what I wanted to do).

Since this is a multi output feedstock, can I keep all the `openfe` and `openfe-base` patches in the same `patch_yaml/openfe.yaml`?

Now the show_diff.py output looks correct:
```
noarch::openfe-base-1.7.0-pyhd8ed1ab_0.conda
+  "constrains": [
+    "dask >=2025"
+  ],
```


Thank you for catching this @hmaarrfk (I kept the build number in (I tested it and it isn't needed) but it also isn't wrong))

PR #1090, opened 2025-10-06
WIP: test impact of marking cudnn builds as broken.

h-vetinari:
In the context of https://github.com/conda-forge/cudnn-feedstock/issues/124, if we end up doing https://github.com/conda-forge/admin-requests/pull/1687, we need to patch existing constraints that would otherwise be unsatisfiable, at least for CUDA 12.x builds.

This is a risky because newer cudnn builds could have added new symbols, and in that case, using an older cudnn at runtime would be a breaking change. However, we may have no other choice if nvidia does not republish a newer cudnn build that supports pre-Turing device code. The hope is that symbol additions are rare enough that this is not a big deal in practice.

However, it appears to be the case that no package depends on these newer cudnn versions:
```
>python show_diff.py
================================================================================
================================================================================
linux-armv7l
================================================================================
================================================================================
win-32
================================================================================
================================================================================
noarch
================================================================================
================================================================================
linux-ppc64le
================================================================================
================================================================================
linux-aarch64
================================================================================
================================================================================
osx-arm64
================================================================================
================================================================================
win-64
================================================================================
================================================================================
linux-64
================================================================================
================================================================================
osx-64
osx-64::mumps-mpi-5.8.1-he6e93e4_4.conda
-    "libgfortran",
+    "libgfortran >=3.0.1,<4.0.0.a0",
```

PR #1107, opened 2025-11-03
Revert "loosen cudnn pins due to builds about to be marked broken"

h-vetinari:
For https://github.com/conda-forge/cudnn-feedstock/issues/124, we originally did #1090 and https://github.com/conda-forge/admin-requests/pull/1687, but the latter was reverted because the hotswap didn't actually work in all cases.

In the meantime, we marked the builds unbroken and [migrated](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7891) for cudnn 9.10, which is almost [finished](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7917).

So we can remove the hotswap, because it'll only lead to misleading (and potentially broken) solves when we redo https://github.com/conda-forge/admin-requests/pull/1687 once the migrator is closed.


h-vetinari:
This is just removing the hotswap, which (as you experienced yourself) did not work. It goes back to the stricter original metadata, and as such I see ~zero risks here.

There's still a minor risk with marking the newer cudnn builds as broken (because the cudnn 9.10 migration will not necessarily have rebuilt _all_ relevant artefacts, just the most recent feedstock state), but that's not something we can avoid anyway.

PR #1109, opened 2025-11-10
Add patch to pvlib and pvlib-python for numpy and scipy dependency

julian-belina:
Checklist


hmaarrfk:
what version is the feedstock at now? has this issue been resolved on the respective feedstocks?

julian-belina:
Both pvlib and pvlib-python are at version 13.1. I don't think they will rebuild old versions. However, I also don't understand how pvlib is built. I can find a feedstock for pvlib-python, but not for pvlib: https://github.com/conda-forge/pvlib-python-feedstock.

PR #1111, opened 2025-11-18
Make published builds of ArcticDB depend on `libcurl <8.17`

jjerphan:
Checklist

See: https://github.com/man-group/ArcticDB/pull/2746

### DIFF
--- a/recipe/patch_yaml/arcticdb.yaml
+++ b/recipe/patch_yaml/arcticdb.yaml
@@ -1,8 +1,9 @@
 # yaml-language-server: $schema=../patch_yaml_model.json
-# Use `libcurl<8.17` dependency for all ArcticDB builds until
-# the regression is understood and fixed. See: https://github.com/man-group/ArcticDB/pull/2746
+# Use `libcurl<8.17` dependency for all ArcticDB builds up to 6.3.1 (inclusive).
+# See: https://github.com/man-group/ArcticDB/pull/2746
 if:
   name: arcticdb
+  timestamp_lt: 1763539708000  # 2025-11-19
 then:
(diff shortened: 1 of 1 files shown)


h-vetinari (changes requested):
> The patch modifies all builds on purpose including the ones in the future.

we don't do that. Add the `<8.17` cap on the feedstock, then patch all _former_ builds (i.e. up to `timestamp_lt:`), and once you have identified the issue and fixed it, you can remove the cap on the feedstock without having to touch the repodata patches again.

recipe/patch_yaml/arcticdb.yaml
```diff
@@ -0,0 +1,10 @@
+# yaml-language-server: $schema=../patch_yaml_model.json
+# Use `libcurl<8.17` dependency for all ArcticDB builds up to 6.3.1 (inclusive).
+# See: https://github.com/man-group/ArcticDB/pull/2746
+if:
+  name: arcticdb
+  timestamp_lt: 1763539708000  # 2025-11-19
```
h-vetinari: Please first add this cap to the arcticdb feedstock. A timestamp here is useless if someone merges another PR on the feedstock that reintroduces a `libcurl <9` bound.

jjerphan: This is being done in the next PR on the feedstock (see https://github.com/conda-forge/arcticdb-feedstock/pull/526#discussion_r2541029761) for the `main` branch, and this has been done for the `rc` branch via https://github.com/conda-forge/arcticdb-feedstock/pull/525.

Do you require it to be done via another, dedicated PR?

PR #1116, opened 2025-11-30
Provide more instantaneous feedback when patching

hmaarrfk:
Generating these patches takes like 6 mins on my wimpy portable laptops. And I can never know if I have a bug in my system.

turns out, a dictionary copy was the longest runtime on my machine...

So i avoid id (and added a comment) then passed around a verbose flag so you can see flashing things on your terminal to give yourself confidence that things are "moving along"


hmaarrfk:
It mostly spews things to the terminal if you have the verbose flag on. So you can be more informed that slow progress is happening.

PR #1119, opened 2025-12-05
Update gen_patch_json.py

tlegier80:
Fix versioning syntax for version 5.0.0

Checklist

### DIFF
--- a/recipe/gen_patch_json.py
+++ b/recipe/gen_patch_json.py
@@ -851,8 +851,8 @@ def _gen_new_index_per_key(index, subdir, index_key="", verbose=False):
         if record_name == "flang":
             deps = record["depends"]
             if record["version"] == "5.0.0":
-                deps += ["clangdev *"]
-                deps += ["flang *"]
+                deps += ["clangdev"]
+                deps += ["flang"]
 
         # add as run_constrained for cling
         if (


tlegier80:
> This will change a bunch of older patches? We need a more targeted approach. Or were the older patches buggy anyways?

I believe the older patches were buggy. The issue that I found was that this version of flang gets entered in conda forge repodata as the following:

```
"flang-5.0.0-he025d50_20180525.tar.bz2": {
        "arch": "x86_64",
        "build": "he025d50_20180525",
        "build_number": 20180525,
        "depends": [
            "clangdev 5.0.0",
            "libflang 5.0.0 h6538335_20180525",
            "openmp 5.0.0",
            "vc >=14,<15.0a0",
            "clangdev * flang*"
        ],
        "license": "Apache 2.0",
        "md5": "6a25fea497e9da30b0aa386db4722fc2",
        "name": "flang",
        "platform": "win",
        "sha256": "7094bc2242e52aea89b8c83e54770028b0668b12e063b405c3423fbfb94f6fa2",
        "size": 2777448,
        "subdir": "win-64",
        "timestamp": 1527899241687,
        "track_features": "flang",
        "version": "5.0.0"
    }
```

The "clangdev * flang*" entry breaks some of conda's python functions because that entry does not contain a valid version. Moreover, it does not appear to be syntactically valid as a single entry

PR #1155, opened 2026-02-14
Add missing `__cuda` to libarrow

h-vetinari:
Fix the environment mix-up in https://github.com/conda-forge/arrow-cpp-feedstock/commit/0544e8701686, for details see https://github.com/conda-forge/arrow-cpp-feedstock/pull/1925

Since this was overlooked for a while, and libarrow gets migrated very often, we need to patch a bunch of artefacts. Still, I think it's beneficial to do this, because we don't want to install CUDA-variants on non-CUDA systems.

Output of `python show_diff.py` in a separate comment due to length restrictions.

### DIFF
--- a/recipe/README.md
+++ b/recipe/README.md
@@ -80,6 +80,8 @@ if:
 
   # any key in the repodata entry (e.g., "version" or "build_number") and a list of values or single value
   <repodata key>_in: <list or single item>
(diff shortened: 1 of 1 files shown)


h-vetinari:
Gentle ping @beckermr

jaimergp (approved):

PR #1157, opened 2026-02-16
Add constraint for urdfdom 3.* and 4.* on urdfdom_headers <2 version

traversaro:
`urdfdom` depends on the `urdfdom_headers` header-only library in its public headers, but urdfdom_headers 2.* removed some headers used by `urdfdom` 4.* and 3.*, this repodata patch ensure that urdfdom 4.* continues to work fine by adding a run_constraint on the urdfdom_headers version.

Checklist


h-vetinari (approved):
Xref also https://github.com/conda-forge/urdfdom-feedstock/pull/46/changes

PR #1196, opened 2026-05-06
raise h5py lower bound on legend-lh5io 0.2.x

bollwyvl:
Checklist

* [ ] Ran `pre-commit run -a` (or `cd recipe && pixi run pre-commit`) and ensured all files pass the linting checks.
  * > don't have time for this to segfault my box this morning, will update with CI results...

ping @conda-forge/legend-lh5io

the upstream pin on h5py [changed between 0.1.0 and 0.2.0](https://github.com/legend-exp/legend-lh5io/compare/v0.1.0...v0.2.0/#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R36).

https://github.com/conda-forge/legend-lh5io-feedstock/pull/3 fixed this for `0.2.3 *_1`, but downstreams might still pull older `0.2.x` builds.

### DIFF
--- a/recipe/patch_yaml/legend-lh5io.yml
+++ b/recipe/patch_yaml/legend-lh5io.yml
@@ -1,0 +1,10 @@
+# yaml-language-server: $schema=../patch_yaml_model.json
+# https://github.com/conda-forge/legend-lh5io-feedstock/pull/3
+if:
+  name: legend-lh5io
+  timestamp_le: 1778077522000
+  version_ge: 0.2.0
+then:
+  - replace_depends:
+      old: h5py >=3.10
+      new: h5py >=3.16
--- a/recipe/patch_yaml/legend-lh5io.yml
+++ b/recipe/patch_yaml/legend-lh5io.yml
@@ -5,6 +5,5 @@ if:
   timestamp_le: 1778077522000
   version_ge: 0.2.0
 then:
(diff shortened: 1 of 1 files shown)


bollwyvl:
Hmm, might not be generally solveable... still investigating.

ManuelHu:
@bollwyvl I am thinking on how we can move this forward. From our side, would publishing  a new legend-lh5io 0.2.4 help, when we move all downstream packages to require `>=0.2.4`? This should hopefully have a similar effect, right? @gipert and me maintain all packages and feedstocks of all downstream packages that we know of, so this should be feasible?

ManuelHu:
@bollwyvl We did some more testing on our side. I am quite confident that there is also another , deeper problem that prevents h5py 3.16 to be installed.

In some tests I could not build a package that additionally pins python==3.10 and h5py==3.16 because something appears to be incompatible and/or not build for this combination. When we bumped the minimum version to python 3.11 the packages build fine again and always get h5py==3.16 installed, despite the broken requirements.

bollwyvl:
Yep, I am only going off what n-levels downstream tests say; I don't _really_ know what all this stuff is doing. It could well be some `numpy` deprecation cycle has lead to some of the dependencies only being built for 3.11+.

PR #1207, opened 2026-06-04
Remove spdlog and fmt from spiceql as run time dependencies

acpaquette:
Checklist


acpaquette:
The `pixi run diff` is giving me timeout issues

PR #1219, opened 2026-06-29
Update urllib import

BrownMunda1:
Checklist

* [ ] Ran `pre-commit run -a` (or `cd recipe && pixi run pre-commit`) and ensured all files pass the linting checks.

### DIFF
--- a/recipe/show_diff.py
+++ b/recipe/show_diff.py
@@ -4,7 +4,7 @@ import difflib
 import json
 import os
 import sys
-import urllib
+import urllib.request
 from concurrent.futures import ProcessPoolExecutor, as_completed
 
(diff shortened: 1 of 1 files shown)


recipe/show_diff.py
```diff
@@ -4,7 +4,7 @@
 import json
 import os
 import sys
-import urllib
+import urllib.request
```
jakirkham, line 7: Seems sensible given this script only uses `urllib.request`. Also better to be explicit about what is being used from `urllib` rather than rely on top-level `import` behavior

jakirkham (approved):
Thanks @BrownMunda1 ! 🙏

PR #1251, opened 2026-09-15
FIX: Add track_features to older dagmc and moab MPI builds

matthewfeickert:
Both feedstocks moved to rattler-build and now use `down_prioritize_variant` to prefer the nompi variant, which adds a track feature (`dagmc-p-0`, `moab-p-0`) to the MPI builds. Older MPI builds were produced by conda-build without any track feature. Since solvers rank track features ahead of version and build number, the old builds now win the solve:

* `'dagmc * mpi_mpich_*'` resolves to 3.2.4 build 3 (mpich 4.3.2, moab tempest build 3) instead of build 7.
* `'moab * mpi_mpich_notempest_*'` resolves to 5.6.0 build 6 instead of build 9.

Add the same track feature to the older builds of the de-prioritized variants so ordering falls back to version and build number.

Assisted-by: ClaudeCode:claude-fable-5-1

---

```console
$ pixi run diff
✨ Pixi task (diff): python show_diff.py  1> show_diff_result.txt: (Slow!! Writes repodata changes to file, add this result to the PR.)
```

[show_diff_result.txt](https://github.com/user-attachments/files/32263804/show_diff_result.txt)

---

Checklist

* [N/A] Only wrote code directly into `recipe/generate_patch_json.py` if absolutely necessary.
  

```
$ pixi run timestamp
✨ Pixi task (timestamp): python -c "import time; print(f'{time.time():.0f}000')": (Generate a UTC timestamp in milliseconds since epoch.)
1789509680000
```


matthewfeickert:
My understanding is that this is needed to be applied to _all_ older releases and not just the builds in the `v3.2.4` series as I _think_ the solver treats track features as a penalty ranked above version and so this applies _everywhere_.

carterbox:
>  which adds a track feature (dagmc-p-0, moab-p-0) to the MPI builds.

Looks like your patch isn't quite right because you're also adding track features to some nompi builds.

e.g.

```
...
osx-64::moab-5.6.0-nompi_tempest_ha812391_3.conda
osx-64::moab-5.6.0-nompi_tempest_ha812391_4.conda
osx-64::moab-5.6.0-nompi_tempest_ha812391_5.conda
osx-64::moab-5.6.0-nompi_tempest_ha812391_6.conda
osx-64::moab-5.6.0-nompi_tempest_hffa3539_0.conda
+  "track_features": "moab-p-0",
```

recipe/patch_yaml/moab.yaml
```diff
@@ -0,0 +1,20 @@
...
+# over the current ones (e.g. `moab * mpi_mpich_notempest_*` resolves to
+# 5.6.0 build 6 instead of build 9). Add the same track feature to the older
+# builds of the de-prioritized variants so ordering falls back to version and
+# build number.
+# c.f. https://github.com/conda-forge/admin-requests/pull/2345#issuecomment-5687428500
+if:
+  name: moab
+  build_in:
+    - mpi_*
+    - nompi_tempest_*
```
matthewfeickert, line 15: @carterbox RE: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/1251#issuecomment-5698906153 the reason why there are `nompi` versions included here is that for [MOAB the `down_prioritize_variant` is](https://github.com/conda-forge/moab-feedstock/blob/547b1e83f5247c0f0367a62a99ada9b66e1a5bbe/recipe/recipe.yaml#L26-L28)

```yaml
  variant:
    # prefer 'nompi' and 'notempest' over other variants
    down_prioritize_variant: ${{ 0 if mpi == "nompi" and tempest == "notempest" else 1 }}
```

so I think the patch is correct as is.

matthewfeickert:
cc @conda-forge/dagmc and @conda-forge/moab

xylar:
@matthewfeickert's moab patches look right to me.  Thank you!

vijaysm (approved):

recipe/patch_yaml/moab.yaml
carterbox: OK! Thanks for confirming.

matthewfeickert: Thanks for checking! This was my first patch submission (I think) and I appreciate people actually reviewing it and helping me out. :pray:

matthewfeickert:
Things are now working as expected in https://github.com/conda-forge/admin-requests/pull/2345#issuecomment-5687428500.

